Issue #140, opened 2015-01-02
Inconsistent Handling of File Format Specification


mandli:
I think the way we decide which file format to try when plotting is not consistent (or at least I cannot see how it works).  When `plotclaw` is called it contains a `format` key word argument which is then used by default and sent through the `plotpages.plotclaw_driver` which does replace it.  Although this is pretty confusing, this actually works save for the spot where it tries to use `plotdata.gettime` which also takes a format argument and uses the default one passed through rather than the one in `plotdata`.  Simple fix to line 1753 of `plotpages` but I am thinking that we need to rethink how we want to specify the file format being used.

mandli:
As an aside, this is also done in https://github.com/clawpack/pyclaw/blob/master/src/pyclaw/plot.py, mainly to adhere to the `plotclaw` interface.

ketch:
In general, I think that file reading is too intertwined with the rest of visclaw.  It's hard to use visclaw to plot a Solution from memory because reading files is such a persistent assumption.  I've done a little to change this, but it would be nice to make the plotting routines not care at all about whether you have a file or something already in memory.

mandli:
I agree, it would be best to abstract this so that the visualization assumed only that it had a `solution` or `state` even.  Perhaps closing this issue and raising another with that design goal would be good.  Another idea I have been having lately is to create a wiki page here on github specifically for a "wish-list" of things we would like to change in VisClaw so we do not rediscover these types of things and they are all in one place.  Perhaps we can start to outline a refactor that could start to be done piece by piece rather than as a whole which at least has been a stumbling block for my attempts.

rjleveque:
This discussion is out of date.  Raise a new issue if there's still a problem.

PR #144, opened 2015-02-20
Fix format handling in `gettime` to be more general

mandli:
Previously if `gettime` was ever called and the format requested was not `petsc` or `ascii` the code would raise an exception.  Now this checks to see if the file format requested's module contains the function `read_t` and uses that.  Since this routine only cares about time the inconsistent number of variable returned by some of the formats (`binary` in this case) is handled gracefully.  If not `read_t` exists then `None` is returned.

Other fixes:
 - Corrected tabs to spaces and some other spacing issues
 - Addressed the file format specification mentioned in #140 as this bug only appears when file formats are properly handled.

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -220,15 +220,19 @@ class ClawPlotData(clawdata.ClawData):
         this routine returns None if that's the case.
         """
         if format == 'petsc':
-            from clawpack.petclaw import io
-            read_t = io.petsc.read_t
-        elif format == 'ascii': 
-            from clawpack.pyclaw import io
-            read_t = io.ascii.read_t
+            import clawpack.petclaw.io
+            read_t = clawpack.petclaw.io.petsc.read_t
 
+        else:
+            module = __import__("clawpack.pyclaw.io.%s" % format, 
+                                                fromlist=['clawpack.pyclaw.io'])
+            if "read_t" in dir(module):
+                read_t = module.read_t
+            else:
+                return None
 
-        t, meqn, npatches, maux, num_dim = read_t(frameno, path=outdir)
-        return t
+        header_info = read_t(frameno, path=outdir)
+        return header_info[0]
 
     def clearfigures(self):
         """
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -81,7 +81,7 @@ class ClawPlotData(clawdata.ClawData):
         self.add_attribute('setplot',False)            # Execute setplot.py in plot routine
 
         self.add_attribute('mapc2p',None)              # function to map computational
-	                                    # points to physical
+                                        # points to physical
 
 
         self.add_attribute('beforeframe',None)         # function called before all plots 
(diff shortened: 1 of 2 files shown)


mandli:
Note that I am not absolutely certain that this is the right thing to do here but seemed like the most obvious solution.

ketch:
It seems to me that the `gettime` function should live in `pyclaw.io`, not in visclaw.

mandli:
That's true and really should be the purpose of the `read_t` functions.  Since this function is not always available I think this was an attempt at getting around that.  Perhaps it would be better to always include even an empty `read_t` although I am not sure if this would work with the PETSc vs. PyClaw IO question.

ketch:
Hmmmm, but why not at least move this function to `src/pyclaw/io/gettime.py`?

ketch:
Of course, if you want to just merge this and perhaps open an issue suggesting that IO stuff like this be migrated in the future, I'm okay with that.

mandli:
I think this makes sense although we would still have to be careful about importing the PETSc IO right?

mandli:
I am all for that.  I think we discussed some of this in #140 as well and just came to the conclusion that VisClaw needs to be disentangled from IO.

ketch:
I think the contents of the function would be exactly the same (i.e., it would still take a format argument).

mandli:
It would be importing `clawpack.petclaw.io` into the namespace of `clawpack.pyclaw.io` but I guess I do not inherently see an issue there.  Should we move this function and see what breaks?

ketch:
Sounds good to me.  

If I understand correctly, the point of this routine is to get the time without reading a whole (possibly huge) outpute file.  I believe there should be a way to get the time from an HDF5 file without reading the whole file, so we should add that at some point.

mandli:
I know we have talked about this before but would something like the following seem reasonable to have as a general purpose object that could handle the common file IO tasks?

```python
class OutputController(object):
    r""""""


    @property
    def file_format(self):
        return self._file_format
    @file_format.setter
    def file_format(self, value):
        if value.lower() == 'petsc':
            self._io_module = __import__("clawpack.petclaw.io.petsc", 
                                   fromlist=["clawpack.petclaw.io"])
            self._file_format = value
        else:
            self._io_module = __import__("clawpack.pyclaw.io.%s" % value, 
                                   fromlist=['clawpack.pyclaw.io'])
            self._file_format = value

        

    def __init__(self, output_path, file_format='ascii'):
        r""""""

        self._io_module = None
        self._file_format = None

        self.output_path = output_path
        self.file_format = file_format


    def get_time(self, frame_num):

        if "read_t" in dir(self._io_module):
            return self._io_module.read_t(frame_num, path=self.output_path)[0]
        else:
            return None
```
I added this to PyClaw while removing (and replacing) the calls to `gettime`.  I am thinking something along these lines could handle what we need to abstract away some of this file format garbage.

ketch:
Yes, that looks good to me.  We can grow this object to let it take over more tasks in the future.  

You can combine the two lines that read

    self._file_format = value

mandli:
I am going to close this PR and open a new one with the proposed changes.  I am not crazy on the name of the module but we can discuss that there.

Issue #118, opened 2014-10-05
colormaps.add_colormaps doesn't always work 


rjleveque:
This is used in geoclaw.topotools for plotting topography and sometimes the break in colors to not match the Z=0 contour.  Example:

```
from clawpack.visclaw import colormaps
from pylab import *

land_cmap = colormaps.make_colormap({ 0.0:[0.1,0.4,0.0],
                                     0.25:[0.0,1.0,0.0],
                                      0.5:[0.8,1.0,0.5],
                                      1.0:[0.8,0.5,0.2]})
sea_cmap = plt.get_cmap('Blues_r')

topo_extent = (-150, 500)

cmap = colormaps.add_colormaps((land_cmap, sea_cmap),
                 data_limits=topo_extent, data_break=0.0)

x = linspace(0,1,11)
X,Y = meshgrid(x,x)
Z = X + Y - 1

pcolor(X,Y,Z,cmap=cmap)
colorbar()

contour(X,Y,Z,[0.])
```

![bad_cmap](https://cloud.githubusercontent.com/assets/720122/4517543/46f502a8-4c41-11e4-8430-4d496bb365b7.png)

mandli:
Slightly modified test code from above
```python
import numpy
import matplotlib.pyplot as plt
import clawpack.visclaw.colormaps as colormaps
land_cmap = colormaps.make_colormap({ 0.0:[0.1,0.4,0.0],
                                     0.25:[0.0,1.0,0.0],
                                      0.5:[0.8,1.0,0.5],
                                      1.0:[0.8,0.5,0.2]})
sea_cmap = plt.get_cmap("Blues_r")
topo_extent = (-150, 500)
cmap, norm = colormaps.add_colormaps((land_cmap, sea_cmap), 
                                     data_limits=topo_extent,
                                     data_break=0.0)
x = numpy.linspace(0,1,11)
X,Y = numpy.meshgrid(x,x)
Z = X + Y - 1

plt.pcolor(X, Y, Z, cmap=cmap, norm=norm)
plt.colorbar()
plt.contour(X, Y, Z, [0.])
plt.show()
```
now produces
![figure_1](https://cloud.githubusercontent.com/assets/322129/10924921/4d137b6c-8258-11e5-80ea-42fa3201eab2.png).

ketch:
Looks like this was fixed.  Closing.

PR #150, opened 2015-03-20
Googleearth

donnaaboise:
Latest Google Earth plotting includes transparency and new colors!  Time slider is fixed!

Thanks Tommaso!!


src/matlab/getlegendinfo.m
```diff
@@ -1,4 +1,4 @@
-function [level_handles,level_labels_out] = getlegendinfo;
+function [level_handles,level_labels_out] = getlegendinfo(base_level);
```
mandli: Does this impact the Google Earth stuff or was this a different bug fix?

src/python/visclaw/buildpykml.py
```diff
@@ -0,0 +1,54 @@
+import pykml
```
mandli: What is this file for?  It does not seem to be used anywhere else.

src/python/visclaw/colormaps.py
```diff
@@ -29,35 +29,41 @@ def make_colormap(color_list):
     html hex string, e.g. "#ff0000" for red.
     """
```
mandli: Can you add documentation to this function pointing out that alpha channels have been added as an optional specification?

src/python/visclaw/data.py
```diff
@@ -223,7 +227,6 @@ def getframe(self,frameno,outdir=None,refresh=False):
             framesoln = self.framesoln_dict[key]
 
         return framesoln
-
```
mandli: Need to add back the white-space here, really should have two lines in between function declarations.

src/python/visclaw/data.py
```diff
@@ -621,12 +622,17 @@ def __init__(self, name, figno, fig_type, plotdata):
         self._show = True
         self.add_attribute('plotaxes_dict', {})
         self.add_attribute('type',fig_type)   # = 'each_frame' or 'each_run' or 'each_gauge'
+        self.add_attribute('use_for_kml',False)
+        self.add_attribute('kml_dpi',200)
+        self.add_attribute('kml_xlimits',None)
```
mandli: As we discussed at the Clawpack developers meeting yesterday and now looking at the code I think `xlimits` and `ylimits` really should be in the `PlotDataAxes` object where they usually are making it more trivial to convert a given figure to one being plotted on Google Earth.

src/python/visclaw/data.py
```diff
@@ -621,12 +622,17 @@ def __init__(self, name, figno, fig_type, plotdata):
         self._show = True
         self.add_attribute('plotaxes_dict', {})
         self.add_attribute('type',fig_type)   # = 'each_frame' or 'each_run' or 'each_gauge'
+        self.add_attribute('use_for_kml',False)
+        self.add_attribute('kml_dpi',200)
```
mandli: DPI is usually set as a dictionary key in the `kwargs` dictionary.  It might be worthwhile to check to see if it has been specified there.

src/python/visclaw/frametools.py
```diff
@@ -176,22 +183,69 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
...
                 # ----------------
-    
+
+                # Mask out covered coarse grid regions.
                 for stateno,state in enumerate(framesoln.states):
                     #print '+++ stateno = ',stateno
-                    state = framesoln.states[stateno]
                     patch = state.patch
 
+
```
mandli: Duplicate statements here.

src/python/visclaw/frametools.py
```diff
@@ -176,22 +183,69 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
...
                     current_data.add_attribute('q',state.q)
                     current_data.add_attribute('aux',state.aux)
                     current_data.add_attribute('xlower',patch.dimensions[0].lower)
                     current_data.add_attribute('xupper',patch.dimensions[0].upper)
-                    
+
                     current_data.add_attribute("x",patch.grid.p_centers[0])
                     current_data.add_attribute("dx",patch.delta[0])
 
+                    # -------------------------------------------------------------
```
mandli: Was this added to support transparency for Google Earth?  If so I am not sure if this is how we would want to go about doing this.  At the very least maybe only adding this to the `current_data` dict if `use_for_kml` is set to `True` or compute it later.

src/python/visclaw/frametools.py
```diff
@@ -247,7 +301,9 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
                     print "*** use ClawPlotAxes.afteraxes "
                     print "*** or  ClawPlotItem.afterpatch instead"
                 try:
-                    if plotitem.has_attribute('add_colorbar') and plotitem.add_colorbar:
+                    if plotfigure.use_for_kml:
```
mandli: Is this written this way to support future use of a colorbar?

src/python/visclaw/frametools.py
```diff
@@ -281,7 +339,26 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
                 plt.axis('image')
 
             # set axes limits:
-            if (plotaxes.xlimits is not None) & (type(plotaxes.xlimits) is not str):
```
mandli: Why not use the `plotaxes.xlimits` at this point?

src/python/visclaw/frametools.py
```diff
@@ -311,14 +386,22 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
...
-                    format=plotdata.print_format, plotdir=plotdata.plotdir,\
-                    verbose=verbose)
+        for figname in plotdata._fignames:
+            plotfigure = plotdata.plotfigure_dict[figname]
+            figno = plotfigure.figno
+            if figno in plotted_fignos:
+                if figno in kml_fignos:
+                    printfig(frameno=frameno, figno=figno, \
+                             format=plotdata.print_format, plotdir=plotdata.plotdir,\
+                             verbose=verbose,kml_fig=True,kml_dpi=plotfigure.kml_dpi)
```
mandli: This relates to a point above, you should not need a separate `kml_dpi` if you rely on the `ClawPlotFigure` `kwargs` attribute, something like `plotfigre.kwargs.get('dpi', 200)` should use a default if needed or use the one in the `kwargs` dictionary.

src/python/visclaw/frametools.py
```diff
@@ -365,7 +448,7 @@ def plotitem1(framesoln, plotitem, current_data, stateno):
 #==================================================================
     """
     Make a 1d plot for a single plot item for the solution in framesoln.
-
+m
```
mandli: Extraneous `m`.

src/python/visclaw/frametools.py
```diff
@@ -1144,17 +1242,6 @@ def printframes(plotdata=None, verbose=True):
     return plotdata
     # end of printframes
```
mandli: I am fine with deleting this at it has been deprecated for awhile but we should check with everyone to make sure that this is really the case.

src/python/visclaw/geoplot.py
```diff
@@ -13,36 +13,45 @@
...
 purple = [0.8,0.3,0.8];
 
+blue_a = [0.0,0.0,1.0,1.0];
+transparent = [0.0, 0.0, 0.0,0.0]
+red_a = [1.0,0.0,0.0,1.0]
+white_a = [1.0,1.0,1.0,1.0]
+
 # Colormaps
 TSUNAMI_MAX_AMPLITUDE = 0.6
+googleearth_colormap = colormaps.make_colormap({-TSUNAMI_MAX_AMPLITUDE:white_a,
```
mandli: We really should not use an effectively sequential colormap for something that should be a divergent one.  See http://matplotlib.org/users/colormaps.html for more info.  I would favor perhaps using your newly defined `blue_a` as the bottom of this colormap.

src/python/visclaw/geoplot.py
```diff
@@ -179,12 +188,18 @@ def surface(current_data):
     q = current_data.q
     h = q[0,:,:]
     eta = q[3,:,:]
-    water = ma.masked_where(h<=drytol, eta)
+
```
mandli: This does not yield the same object that was previously returned and needs to be made consistent again.

src/python/visclaw/geoplot.py
```diff
@@ -196,15 +211,25 @@ def surface_or_depth(current_data):
...
-    surface = ma.masked_where(h<=drytol, eta)
-    depth = ma.masked_where(h<=drytol, h)
-    surface_or_depth = where(topo<0, surface, depth)
+
+    # With this version, the land was plotted as white in png files for KML.
+    # surface = ma.masked_where(h <= drytol, eta)
+    # depth = ma.masked_where(h <= drytol, h)
+    # surface_or_depth = where(topo<0, surface, depth)
+
+    # With this version, the land is transparent.
```
mandli: Again I do not think this yields the same masking as before.

src/python/visclaw/plotclaw.py
```diff
@@ -48,7 +48,7 @@ def plotclaw(outdir='.', plotdir='_plots', setplot = 'setplot.py',
     plotdata.format = format
     plotdata.msgfile = msgfile
 
-    plotpages.plotclaw_driver(plotdata, verbose=False, format=format)
```
mandli: Should maintain this white space.

src/python/visclaw/plotpages.py
```diff
@@ -5,8 +5,6 @@
 Utilities for taking a set of plot files and creating a set of html and/or
 latex/pdf pages displaying the plots.
 """
-
```
mandli: Should maintain this whitespace.

src/python/visclaw/plotpages.py
```diff
@@ -62,11 +60,10 @@ def __init__(self):
         self.timeframes_fignos = 'all'
         self.timeframes_fignames = {}
         self.timeframes_prefix = 'frame'
-        
-    
-        self.pageitem_list = []
```
mandli: This white-space should be cleaned up but again we should have two lines between function definitions.

src/python/visclaw/plotpages.py
```diff
@@ -549,7 +545,242 @@ def plots2latex(plot_pages_data):
...
+        try:
+            # If kml_starttime is already time_struct, we won't be able to
+            # extend it this try block will fail
+            user_time = plotfigure.kml_starttime      # 6 entries of a 9-tuple
+            user_time.extend([0,0,-1])                # Extend to 9 tuple
+        except:
+            user_time = time.gmtime()                 # 9-tuple local time
+
+        # This doesnt' really do the right thing.
+        starttime = time.mktime(user_time)
```
mandli: Would `datetime` work better for you here?  I think it might do what you want.

src/python/visclaw/plotpages.py
```diff
@@ -549,7 +545,242 @@ def plots2latex(plot_pages_data):
...
+            arg_list = ["gdal2tiles.py", \
+                       "--profile=geodetic", \
+                       "--force-kml", \
+                       "--resampling=near", \
+                       "%s.vrt" % (fname_str)]
+            retval = retval or subprocess.call(arg_list)
+
+            if retval > 0:
+                print "gdal : something went wrong!\n"
+                sys.exit(1)
```
mandli: Probably better to raise an appropriate exception here rather than just exiting.

src/python/visclaw/plotpages.py
donnaaboise: I’ve actually fixed the time issue - will push that shortly.

src/python/visclaw/plotpages.py
```diff
@@ -549,7 +545,242 @@ def plots2latex(plot_pages_data):
...
+            zip.write("%s.vrt" % (fname_str))
+
+            # Add the <fname>/ directory
+            for dirname, subdirs, files in os.walk(fname_str):
+                zip.write(dirname)
+                for filename in files:
+                    zip.write(os.path.join(dirname, filename))
+
+            # Clean up files
+            os.remove("%s_tmp.vrt" % (fname_str))
```
mandli: Safer thing to do here if you really need temporary files is to use Python's built-in package for them `tempfile`.

src/python/visclaw/sample.py
```diff
@@ -0,0 +1,58 @@
+from lxml import etree
```
mandli: Is this file needed?

src/python/visclaw/testfile.kml
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
```
mandli: Do you want to evolve this into an actual test?

mandli:
Sorry for all the comments, it looks good overall save for some concern about changing the `plot_var` functions and some other odds and ends.  The whitespace changes I think are mostly fine but there were a couple spots that it was a bit too aggressive I think.  Also the file `ipyclaw.py` should remain executable.

mandli:
BTW, can you provide a pointer to an example?

src/python/visclaw/buildpykml.py
donnacalhoun: we are not using this file - it can be deleted.

rjleveque:
Yes, it would be nice to provide a sample or two of how you use this, e.g. a setplot.py to use with the chile2010 example.  

You might want to start a tsunami/googleearth directory in the apps repository with some examples.

mandli:
I would say just have another setplot.py (maybe `setplot_googleearth.py`) in the same directory, otherwise we will have to copy everything unless it truly will be a different example.  Also we could add something to the apps repository.

src/python/visclaw/data.py
donnaaboise: I guess I had in mind that this dpi setting would be specific to the KML output, and not applied to other figures generated for html output, for example.    

But then I am not really sure where this kwargs gets set.    What did you have in mind?

src/python/visclaw/frametools.py
donnaaboise: I didn't add this;   our earliest commit on the masking was 1e5d8dc63  All of the 'add_attribute' lines were already there at this point.

src/python/visclaw/frametools.py
donnaaboise: But would this then mean that all figures have the same dpi?  I think 200 is probably fine for figures that will be visualized using the web interface.   But for GE, the figures are tiled, and so a much higher dpi is required - 1600?   Since these figures are slower to generate, I wouldn't impose this dpi on all other figures.

src/python/visclaw/geoplot.py
donnaaboise: That was the point - the masking before didn't work for GE.    Is this new masking a problem for standard output?

src/python/visclaw/sample.py
donnaaboise: No - not needed.  I removed it.

src/python/visclaw/testfile.kml
donnaaboise: No - I've removed it.

src/python/visclaw/data.py
donnaaboise: We thought about this, but relying on one of the axes limits means that (1) we would have to loop over axes in plotpages, where normally, the lowest level object is a frame, and (2) that we would correctly "guess" the correct axes to choose for this (the first one? ) .  By making it  a figure attribute, we are enforcing the idea that there is only one set of axis limits that can be used when creating plots suitable for visualization in GE.

mandli: I see your point on this although I am not entirely unconvinced that adding new attributes to the figure rather than using those already provided in `axes` is a good idea.  I would suggest that you just take the first axes in a figure and perhaps print a warning that none of the other axes will be used.

An advantage to keeping the interface as similar to the existing one is that someone could switch between normal (boring) plotting and Google Earth visualization very easily and less error prone.  I also think maintaining the structure of the data is really important even though some of the functionality is not possible with the Google Earth stuff.

rjleveque: +1 to keeping `xlimits` associated with the `ClawPlotAxes` if possible.

src/python/visclaw/frametools.py
mandli: I see your point, since the DPI really needs to be higher and that is an important attribute than I agree with you on this.

src/python/visclaw/data.py
donnaaboise: I am still reluctant to rely on the plotaxes settings to guess the GE axes limits.  

What would happen if the user  sets all axis limits to "auto" (or doesn't set anything).  Are there routines in Matplotlib that would allow us to get the guess axes limits?   And is it always the case that the user would use the same axis limits for possibly multiple axes per figure?  If we guess limits for the wrong set of axes, the user has to recreate all of the figures.  Also, it would seem to me that the user would want to be able to specify exactly where GE should place their figure, even if plotaxes is set to 'auto'.  

Maybe one solution is to rename this attribute to something like "kml_latlong_box" to indicate exactly where the frame will be placed in the GE browser.

mjberger: I have using the auto for matplotlib, so I would hope there would be a separate set of attributes for the google earth box, instead of getting rid of it.

mandli: But would you want to do this using a different interface than what is already available?

In response to Donna, you do have a point, it is actually non-trivial to get the axes limits from the automatically generated bounds (you would essential have to fake plot once and grab them afterwards).  That being said, you could always look at the solution object itself when actually plotting and use those as the bounding limits.  My view on the latter problem is that if the user wants to place the axes specifically they should not specify `auto` anyway.  Also, if there are multiple axes per figure we could raise an exception and not allow that to occur.  I think I would still prefer this route, providing feedback but staying with the same interface, than adding similar options but to a different level of the data hierarchy.

donnaaboise: Maybe I should clarify what I am doing with the kml_xlimits and kml_ylimits figure attributes.  In frametools,  if plotfigure.use_for_kml=True, any plotaxes limits (auto,None, or explicitly set) are overridden by  kml_xlimits/ylimits.   This enforces consistency between the figure axes and the latlong box (created in plotpages and also specified by kml_xlimits/ylimits)  that will be used by GE.

If use_for_kml=False, then the user defined plotaxes limits are used in the usual way.  

I could, for example, not override the plotaxes.xlimits/ylimits settings and just go with whatever the user sets (auto, None, etc).   And use the kml xlimits/ylimits only in creating the kml files in plotpages (and possibly change the name of the attribute).     But then if Matplotlib happens to choose a different set of axes than what is specified in the kml settings, the user would see weird distortions in their plots on the GE browser.   

When creating the KML files (which needs a specified Lat/long box), I could guess at what the user had in mind for axes, but then the other issues of multiple sets of axes, and where to figure out what axes to use comes up.

mandli: I guess maybe my contention is that I do not see the advantage of creating a new data item that effectively controls the same plot attribute.  I recognize that the scheme being proposed will not interfere with normal plotting but I think it would be confusing for the user to have new attributes that had `xlimits` and `ylimits` in the names attached to a different object.  I also see a lot of value in being able to switch on and off `use_for_kml` and not change anything from my previous plots and get Google Earth visualization out of it (now come to think of it, why is there a figure type `kml_figure` and this attribute, don't these imply the same thing?)

donnaaboise: I agree that maybe "kml_xlimits" and "`kml_ylimits" are not the best choice of names - it could be confused with the plotaxes attributes.  

As for the name 'kml_figure',  the user can of course name this figure anything they want.  But I did have in mind that a user would use a figure solely for GE plots and not just switch on or off the use_for_kml setting.  That said, setting "use_for_kml=False" should restore a plot to normal settings.

Also, if a figure is to used for GE, then titles, colorbars, tick marks, labels are ignored.   In fact, there are only very few things that the user can really set for a GE plot, so I didn't think it wouldn't be too much extra code to have a dedicated GE figure (hence the name "kml_figure").   That said, the GE figures all show up in the usual HTML output.  I am not sure this is so helpful, other than to verify that the figures look okay.

donnaaboise: I just tried plotting by relying only on the 'auto' selection mode, and half way through the simulation. Matplotlib changes its mind about what the axes limits should be.  Early figures match the GE latlong box (hardwired with kml_xlimts/ylimts), and later ones don't.     

So I guess one question is, should the GE box be allowed to  change during the simulation?  (I say no, at least in this version).

mandli: Ah you are right, it's just a name so that's not so useful.  You also make another good point that GE plots ignore a lot of the options and I agree that it is still important to have them around so you can switch on and off the GE plots.

mandli: As to the `auto` problem, I am a bit disturbed that matplotlib changed its mind half way through.  How different is it?  For the record, I don't think this should happen regardless as the full domain of the solution should always be completely plotted, I guess it could be adding more buffer between the plot and the data?

donnaaboise: I have not idea why Matplotlib changes its mind - I think the data is still the same, but it adds padding at the top of the plot (effectively adding something to the upper axis limit, in this case).   It could just be a round off thing.    In any case, I don't think we can rely on 'auto' to do the right thing.

It is really different - at least 5 degrees?  

frame0010fig1

frame0011fig1

mandli: Hmm, that is strange.  Would raising an exception if `use_for_kml == True` and `xlimits` and `ylimits` are not set work then?

donnaaboise: Sure - that is what I do now if #kml_xlimits/ylimits" are not set.  

The question then, is, which axes would we grab the xlimts/ylimits from, and how does this get communicated back to plotpages (where the kml files are created).    It seems to make sense to grab the axes information from the type "2d_color" plot. 

Ultimately, what we'd want to avoid is a situation where the plots that are generated don't line up exactly with the GE topography, but that the user is left with no obvious reason why.  Since they already have to set several things in the plotfigure attributes, one more item (x/y limits) didn't seem like such a big deal.

donnaaboise:
Would you prefer completely separate directories and/or setplot.py files for a GE example?  Or is it enough to include some GE switches (use_for_kml, for example) in an existing setplot.py file? I am fine either way.

mjberger:
I’d be fine either way. Not every example is suitable for GE, so I guess having separate files would make it clear which ones are or aren’t.

Marsha

src/python/visclaw/data.py
mandli: Just allowing these to be set the usual way.  I think I see your point about there being a difference so I guess I would be in favor of keeping it the way you have it now.

src/python/visclaw/frametools.py
mandli: I am trying to track this down and perform some tests, sorry to blame you guys!

src/python/visclaw/geoplot.py
mandli: Not sure yet, see above.

rjleveque:
If kml plots require a fairly different set of attributes than ordinary plots, would it make more sense to define a new class `ClawKmlFigure` to use rather than `ClawPlotFIgure`?  So in the same `setplot.py` you might have something like

```
plotfigure = plotdata.new_plotfigure(...)
# set things for an ordinary figure if desired

kmlfigure = plotdata.new_kmlfigure()
# set other things for a kml figure
```

where `new_kmlfigure` would return an object of class `ClawKmlFigure`?

This goes a bit against the idea that different figures might use different backends, e.g. `matplotlib` or `paraview`, so another possibility is 

```
plotfigure = plotdata.new_plotfigure(...)
plotfigure.backend = "matplotlib"
# set things for an ordinary figure if desired

plotfigure = plotdata.new_plotfigure()
plotfigure.backend = "kml"
# set other things for a kml figure
```

which I guess is the current system with `use_for_kml` rather than `backend`.  

At any rate it seems ok to me if different backends require somewhat different attributes, although I agree it would be best to avoid confusion with similar names attached to different levels.

But I think we should make it possible to specify two different types of plots in the same `setplot.py` since you might want `make plots` to produce both a set of plots in lat-long rectangles with matplotlib and a set of kml files.  

Also, before over-thinking this, does the flatter format propsed by @ketch for griddle make this any easier, and are we going to have to rethink this at any rate if we move in that direction?

donnaaboise:
(This is on response to Randy's idea about adding a new plotfigure type).    

I am not quite sure I fully follow this, as I don't see the KML figures as requiring a new backend.    Behind the scenes, matplotlib is still being used to produce the figures.     I am not sure what griddle does, but if it is another way to produce PNG file (or any image format, really) than these could also be used as figures suitable for visualization in GE. 

But what I realize now is that we can really take advantage of is this idea of GE "layers".  Each plotitem (axes?) can be a layer in the GE visualization.   For example, multiple pcolor plots could be specified, each one in its own layer, patch borders could be another layer,  another for a contour plots, etc.  These could be toggled on and off in the GE browser  (basically the way things are now set up with gauges, regions, etc).  The title of each layer could be pulled from the name the user gives to the plotitem ("Sea surface height", "bathymetry", etc).  I would still vote for a global axes limits, possibly still as a plotfigure attribute, but other things, such as colormaps, colorbars. resolution (dpi), tiling,   could be specific to a layer.

donnaaboise:
Maybe I should add that by "layers", I don't think I mean layers in the way GE uses the term.   I mean the more abstract idea of grouping different types of plots into folders, under the Places setting, and then selectively setting the visibility of each "layer".

mjberger:
i like that idea a lot - gives maximum flexibility without having to rerun the plotting all the time.

donnaaboise:
Something like this could be done now, with multiple KML figures (i.e. figures for which `use_for_kml=True`).    Instead of a KMZ file for each figure, as I do now, I would create a single KMZ file, and for each figure, open a new folder in the top level  `doc.kml` file.  
Also, patch borders could be created directly in GE using polygons, with different colors for each level, for example.  Something like what is now done for regions.
This could eliminate the need for a figure attribute that controls, for example, the xlimits.  While a separate colorbar would still have to be created (unless I can figure out how to grab the colorbar object as it is created now), but this would live with the layer.  
Also, users might want to have layers that show different zoom levels (although this is what the tilting does).

rjleveque:
I like this idea, and it seems like you might want to support both multiple `plotaxes` and multiple `plotitems` on each, as VisClaw now does.  For example you might want the pcolor map as one plotitem and contours of bathymetry as another plotitem.  You might also want one plotaxes centered around one location on the coast with data coming from one GeoClaw run and another plotaxes some place else on the coast from a different run (for the same tsunami), with the ability for the viewer to toggle each on or off separately (useful if they are both high resolution, perhaps).  To me this argues to keep axis limits as attributes of plotaxes, not plotfigure.

donnaaboise:
It makes sense to support different axes - at the very least, axes should be a different layer.   This will require that a new PNG file gets created for each axes, and so would require modifications to frame tools.   As for different plotitems, I am not so sure.    If they can all be turned on or off, would there be a reason to have more than one item per axes?  Or would the idea be to create a separate PNG file for each plotitem as well?   At some point, then, the figure/axes/plotitem distinction might start to seem a bit arbitrary, if everything can be done independently.   But I haven't thought about this too much.

donnaaboise:
I now create "levels.kml" which plots patch borders at different levels.  Please test it at the branch "ge_folders" (I know, the name doesn't make much sense).  
screen shot 2015-04-21 at 9 58 30 pm

The good news is that levels can be selectively turned on and off.  The bad news is that the colormap for creating to pcolor plot has to be transparent, or the polygons don't show up.   This seems to be a general problem with GE - there is no way to plot polygons on top of a ground overlay.

mandli:
The function `add_colormaps` (in colormaps.py) is broken now that an alpha channel is included with this PR.

donnacalhoun:
In looking at add_colormaps, it looks like we need to add an alpha channel to the line : 

     new_dict = dict(red=[], green=[], blue=[])

Is there more? 



where is add_colormaps called? I only see it called from topotools.

mandli:
`add_colormaps` is called from `topotools` in the main code but there are a number of places where it is called in app code and could possibly be called from user code.  My guess is this should be a simple modification to `add_colorbar` to include the `alpha` channel if specified or give a default of 1 otherwise.

donnacalhoun:
(did you mean `add_colormaps`, not `add_colorbar`?)

I added an alpha channel to add_colormaps, and tested it with : 

    from clawpack.visclaw import colormaps, geoplot
    cmaps = (geoplot.googleearth_transparent, geoplot.tsunami_colormap)
    new_cmap = colormaps.add_colormaps(cmaps, data_limits=[-0.2,0.2],
                                   data_break=0.0)

Seems to work, but then I am not sure how to test it further.

mandli:
Ah yes, `add_colormaps`. I will test it out.

mandli:
I am not sure what to make of this but I tried out your `add_colormaps` fix and ran into the following exception and traceback
```python
Exception in Tkinter callback
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-tk/Tkinter.py", line 1410, in __call__
    return self.func(*args)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/backends/backend_tkagg.py", line 276, in resize
    self.show()
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/backends/backend_tkagg.py", line 348, in draw
    FigureCanvasAgg.draw(self)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/backends/backend_agg.py", line 451, in draw
    self.figure.draw(self.renderer)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/artist.py", line 54, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/figure.py", line 1034, in draw
    func(*args)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/artist.py", line 54, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/axes.py", line 2086, in draw
    a.draw(renderer)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/artist.py", line 54, in draw_wrapper
    draw(artist, renderer, *args, **kwargs)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/image.py", line 364, in draw
    im = self.make_image(renderer.get_image_magnification())
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/image.py", line 588, in make_image
    transformed_viewLim)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/image.py", line 204, in _get_unsampled_image
    x = self.to_rgba(self._A, bytes=False)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/cm.py", line 261, in to_rgba
    x = self.cmap(x, alpha=alpha, bytes=bytes)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/colors.py", line 542, in __call__
    self._init()
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/colors.py", line 731, in _init
    self.N, self._segmentdata['alpha'], 1)
  File "/Users/mandli/src/envs/claw/lib/python2.7/site-packages/matplotlib-1.4.x-py2.7-macosx-10.8-intel.egg/matplotlib/colors.py", line 458, in makeMappingArray
    "data mapping points must start with x=0. and end with x=1")
ValueError: data mapping points must start with x=0. and end with x=1
```

It is reproducible by running the following commands
```
In [1]: import clawpack.geoclaw.topotools as tt

In [2]: topo = tt.Topography("NewOrleans_3s.tt3")

In [3]: topo.plot()
Out[3]: <matplotlib.axes.AxesSubplot at 0x10d9fbc10>

In [4]: import matplotlib.pyplot as plt

In [5]: plt.show()
```
Here I used some bathymetry I had laying around but it seems to happen with a few of the bathymetries I have laying around.  I also confirmed that it does not occur with the current master.

donnacalhoun:
Hm.. I reproduced this error, more or less (I tried it on your gulf_caribbean.tt3 topo).  I'll see what I can dig up.

mandli:
I just made a PR to your branch @donnaaboise that addresses most of the rest of the concerns labeled above.  From some private conversation I think we should merge this once we finally address the following concerns:

 - [ ] Where `xlimits` and `ylimits` go.  It sounds like there has almost been a consensus but we should move that discussion here and say why the conclusion was made the way it was.
 - [ ] The above `add_colormaps` problem

Is there anything else left hanging at this point?

src/python/visclaw/geoplot.py
mandli: I think I am satisfied that this is actually probably a better way to go in the end and I have not seen any problems.

donnacalhoun:
What is the "almost" consensus on the axes limits issue?

mandli:
Heh, I had hoped that you would know.  I still favor keeping the limits in the axes but I also see your point on the matter.  I was wondering if perhaps we could address your concerns by adding an additional attribute to the axes object that indicates that it is the axes in a figure to use for google earth in case there are multiple (`use_for_kml` for instance)?

donnacalhoun:
Did we decide that a GE figure would only be used for GE plots?  

Or maybe something like this would be okay : 

```python
plotfigure.use_for_kml = True

if (plotfigure.use_for_kml):
    plotfigure.kml_xlimits = [...]
    plotfigure.kml_ylimits = [...]
    # set the colorbar
    # few other KML items
    # no need to explicitly set xlimits, limits in an axes object. 
else:
   # set things as before
   plotaxes.xlimits = [...]
   plotaxes.ylimits = [...]
   
   # plot land
   # plot bathy contours ...
```

mandli:
I guess that would work, is the argument for following this pattern that the `xlimits` and `ylimits` are always going to be different between the two?  What I was thinking was that someone could write basically the same code and the KML pieces would be used if `use_for_kml` was `True`.  To solve the axes problem I was thinking one could tag an axes, maybe something like `plotaxes.use_for_kml`.  If multiple were specified then you can raise an exception.  Alternatively one could specify multiple and treat them as multiple axes.  I think this would require the fewest number of additions to a `setplot` function to work with the Google Earth capabilities and both the regular plotting and the Google Earth plotting could live side-by-side.

donnacalhoun:
What I am hoping to avoid is searching through the plotaxes in both frametools and plotpages (where the KML file gets created).  So  whatever decision gets made in frametools regarding which axes to use, the same decision should be communicated to plotpages.   

With that in mind, I suggest the following : 

```python
plotdata.kml_latitude = [...]
plotdata.kml_longitude = [...]
```
and then let the user specify their plot axes however they want (multiple axes, even).  The user just needs to be aware that if they choose axes that are not consistent with (read : "same as") the LatLong box (above) that is written to the KML file in plotpages, things won't look right in GE.    The axes limits will have to be set somewhere ('auto' won't work), so I am not sure it is such a big deal to require that they be set explicitly for GE, in addition to any plotaxes settings.     

Currently, I override any plotaxes settings, so in fact, the user only has to specify the axes limits as above.   If 'use_for_kml = False`, then any auto settings would be used. 

In any case, are  the axes the sort of thing that a user will change often?

donnacalhoun:
I am looking at the colormap issue, and am wondering what the context is for this call to topo.plot(). 
It seems that in normal plotting, everything works.  So is there something special that happens in topo.plot()?  I can reproduce the bug as above, but am wondering about the pathways under typical operating conditions that lead to this error.

mandli:
I use it a lot from a python prompt to easily get a plot of a topography file although it's also useful as an "other figure" for GeoClaw output.  Could it be the particular color maps that are being joined?  Also note that the `add_colormaps` function does not entirely work yet #118.

mandli:
More simple example of `add_colormaps` breaking
```python
#!/usr/bin/env python

import numpy
import matplotlib.pyplot as plt

import clawpack.visclaw.colormaps as colormaps

land_cmap = colormaps.make_colormap({ 0.0:[0.1,0.4,0.0],
                                     0.25:[0.0,1.0,0.0],
                                      0.5:[0.8,1.0,0.5],
                                      1.0:[0.8,0.5,0.2]})
sea_cmap = plt.get_cmap('Blues_r')
cmap = colormaps.add_colormaps((land_cmap, sea_cmap), 
                               data_limits=[-1,1],
                               data_break=0.0)

X,Y = numpy.meshgrid(numpy.linspace(-1,1,100),numpy.linspace(-1,1,100))

fig = plt.figure()
axes = fig.add_subplot(1, 1, 1)
plot = axes.pcolor(X, Y, X*Y, cmap=cmap)
fig.colorbar(plot)

plt.show()
```

donnacalhoun:
I think I fixed the colormap issue - Kyle's test (above) should work now.

mandli:
Works for me!

If you merge in the PR on your branch from me and update this PR to pull in the newest code so that we can merge this I think we are good to go.

rjleveque:
Great!  `Topography.plot` works for me now too.

But it looks like some conflicts need to be resolved.

@donnaaboise:  It would be great to also create some documentation, e.g. a file `clawpack/doc/doc/kml.rst` that gives at least a summary of what's available and points to an example. It would be good to list the new dependencies too.  (By the way I did get `gdal` working with `brew`, so tiling now works for me too.)

I guess there should also be a PR on `clawpack/geoclaw` to include such an example, or did you plan to put it in `clawpack/apps`?

donnacalhoun:
I have started documentation  (see donnacalhoun/googleearth/doc) and plan to include the things you've mentioned.  

And yes, I need to issue a PR on related geoclaw stuff - so far, there is only the chile2010 example, and a few modifcations to kmltools and geoplot.

I'll merge in latest master into my GE branch - hopefully, that should take care of any conflicts.

mandli:
I just noticed that the KML page seems to always be created regardless of whether it is needed, is this intended behavior?

donnaaboise:
Do you mean  the actual .kml file?  This is created if “plotfigure.kml_publish” is set to something.   If this is not 
set (i.e. default = None) or set to None, this .kml file will not be created.   Look in the setplot.py file - if you have one of the google earth files I wrote, you probably have a URL set - this will be removed here shortly.

This file just contains links to a KMZ file on a remote server, and can be easily emailed or downloaded.   The KMZ files
can be quite large, if there are lots of frames/figures.

donnacalhoun:
I did just notice that the attribute `plotdata.kml`  (analogous to `plotdata.html`) was set to `True` by default.  Maybe this is what you meant?   I've switched it to `False` by default.

mandli:
Yeah, I think that's what I was seeing.  I would not have noticed except I had this branch checked out but not the GeoClaw changes and so none of the plotting was working (kmltools were not updated).

donnacalhoun:
I just pushed the geoclaw examples - the chile2010 and ike examples should work (under ike, see setplot_ge.py).

rjleveque:
@donnaaboise: You said  (see donnacalhoun/googleearth/doc).  Where is this?

donnacalhoun:
Start of documentation is [here](https://github.com/donnacalhoun/doc/blob/googleearth/doc/googleearth_plotting.rst)

Still very much WIP ... so not really really for commenting yet.

donnacalhoun:
I just tested the Google earth stuff on Linux (Ubuntu) and it works great.   In fact, I realize now that a conda install of GDAL is the way to go : 

          % conda install gdal
          % export GDAL_DATA=$ANACONDA/share/gdal

I also upgraded my Mac version, and that also works.  So I think I will recommend the conda install in the documentation.  Continuum must have just started including these support files with the installation.

If anyone else wants to try this to confirm, that would be great.     Just check that `gdal-config --datadir` is 
returning the `anaconda/share/gdal` directory.

rjleveque:
I managed to install gdal using brew,  so I haven't tried this.

rjleveque:
In general this is all working fine for me now.  But I did just try it on a directory with 50 frames of a simulation over the entire Pacific and ran into problems creating the zip file:

```
...
  File "/Users/rjl/git/clawpack/clawpack/visclaw/plotpages.py", line 916, in plotclaw2kml
    zip.write(os.path.join(dirname, filename))
  File "/Users/rjl/anaconda/lib/python2.7/zipfile.py", line 1146, in write
    self._writecheck(zinfo)
  File "/Users/rjl/anaconda/lib/python2.7/zipfile.py", line 1114, in _writecheck
    " would require ZIP64 extensions")

zipfile.LargeZipFile: Files count would require ZIP64 extensions
```

donnacalhoun:
What is the resolution you are using?  In general, I found it is possible to go with a much lower resolution than what I had earlier thought.  But it is important to keep the dpi consistent with the AMR grids.  

I am not sure what this ZIP64 extension is, but I can look into it also.   At the very least, we should raise a Large file exception.

mjberger:
brew want to take over and clean/fix your entire file system.  Installing via conda is an easier proposition.

rjleveque:
Not sure why I got that zipfile error, I think I had the resolution set much too fine for the full Pacific simulation.   

Things seem to work great now, I'll go ahead and merge.

PR #151, opened 2015-03-27
Added more compatibility with 3d gauges.

cjvogl:
Added num_dim and additional printing code to account for 3rd dimension when using 3d gauges.

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -25,7 +25,7 @@ class ClawPlotData(clawdata.ClawData):
     """
 
     # ========== Initialization routine ======================================
-    def __init__(self, controller=None, num_dim=2):
+    def __init__(self, controller=None):
         """Initialize a PlotData object
         
         """
@@ -48,8 +48,6 @@ class ClawPlotData(clawdata.ClawData):
             self.add_attribute('rundir',os.getcwd())     # uses *.data from rundir
             self.add_attribute('outdir',os.getcwd())     # where to find fort.* files
             self.add_attribute('format','ascii')
-
-        self.add_attribute('num_dim',num_dim)
 
         # This should eventually replace all need for recording the above
         # information
@@ -291,7 +289,7 @@ class ClawPlotData(clawdata.ClawData):
                     print "continuing..."
                     return None
                     
-                gauge_data = amrclaw.GaugeData(self.num_dim)
+                gauge_data = amrclaw.GaugeData()
                 gauge_data.read(outdir)
 
                 # Check to make sure the gauge requested is in the data file
@@ -302,7 +300,7 @@ class ClawPlotData(clawdata.ClawData):
                 # solutions below
                 locations = {}
                 for gauge in gauge_data.gauges:
-                    locations[gauge[0]] = gauge[1:1+self.num_dim]
+                    locations[gauge[0]] = gauge[1:-2]
 
             except:
                 if verbose:
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -25,7 +25,7 @@ class ClawPlotData(clawdata.ClawData):
     """
 
     # ========== Initialization routine ======================================
-    def __init__(self, controller=None):
+    def __init__(self, controller=None, num_dim=2):
         """Initialize a PlotData object
         
         """
@@ -48,6 +48,8 @@ class ClawPlotData(clawdata.ClawData):
             self.add_attribute('rundir',os.getcwd())     # uses *.data from rundir
             self.add_attribute('outdir',os.getcwd())     # where to find fort.* files
             self.add_attribute('format','ascii')
+
+        self.add_attribute('num_dim',num_dim)
 
         # This should eventually replace all need for recording the above
         # information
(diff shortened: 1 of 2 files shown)


src/python/visclaw/gaugetools.py
```diff
@@ -102,6 +102,8 @@ def __repr__(self):
             output = "%4i" % self.number
             output = " ".join((output,"%19.10e" % self.location[0]))
```
mandli: This could probably be cleaned up using either a list comprehension or a for loop.

src/python/visclaw/data.py
```diff
@@ -25,7 +25,7 @@ class ClawPlotData(clawdata.ClawData):
     """
 
     # ========== Initialization routine ======================================
-    def __init__(self, controller=None):
+    def __init__(self, controller=None, num_dim=2):
```
mandli: This is going to break a lot of `setplot.py` scripts!

mandli:
Is there anyway to avoid adding `num_dim` to the constructor?

src/python/visclaw/data.py
```diff
@@ -300,7 +302,7 @@ def getgauge(self, gaugeno, outdir=None, verbose=True):
                 # solutions below
                 locations = {}
                 for gauge in gauge_data.gauges:
-                    locations[gauge[0]] = gauge[1:3]
+                    locations[gauge[0]] = gauge[1:1+self.num_dim]
```
rjleveque: I think we could avoid `num_dim` by changing this to 
       `locations[gauge[0]] = gauge[1:-2]`
since the location `x,y` or `x,y,z` is always followed by exactly two values `t1` and `t2`.

src/python/visclaw/data.py
cjvogl: I thought by specifying a "default" value for num_dim, all the existing init calls would still work?  My python experience is pretty limited... so I'm probably missing something.  Looking at the rest of the comments, this should be possible without adding the attribute.

src/python/visclaw/data.py
cjvogl: I'll make this change and remove the num_dim attribute

src/python/visclaw/gaugetools.py
cjvogl: Will do.

src/python/visclaw/data.py
```diff
@@ -289,7 +291,7 @@ def getgauge(self, gaugeno, outdir=None, verbose=True):
                     print "continuing..."
                     return None
                     
-                gauge_data = amrclaw.GaugeData()
+                gauge_data = amrclaw.GaugeData(self.num_dim)
                 gauge_data.read(outdir)
```
cjvogl: Not sure how to get arorund using num_dim here (in order to pass this info along to GaugeData in amrclaw.

mandli: Can we maybe have `num_dim` be an argument to `getgauge` or is that really not helping matters?

src/python/visclaw/data.py
mandli: True but for anything that uses `num_dim` and is not 2D this will break.  Since nothing really uses `num_dim` maybe this won't be as bad as I was first suggesting but the parameter will be set incorrectly everywhere but for 2D runs.

cjvogl: True.  Seems to me the best way to approach this is your suggestion to move num_dim as an argument to getgauge, with a default value of 2.  Granted, this still won't be correct for non-2d runs... but at least it will only be wrong in getgauge (and GaugeData has it's own num_dim attribute that has a default value of 2d anyways).

src/python/visclaw/data.py
cjvogl: I think this is the way to go (see my "other" comment... I really need to learn to use the referencing and syntax highlighting here).

rjleveque:
`getgauge` doesn't need to know `num_dim` for reading `fort.gauges` since this always has the same format regardless of the number of dimensions.   We only need `num_dim` for reading or printing the gauge locations and it seems like it should be possible to simplify things if we refactor a bit, and use things like `gauge[1:-2]`.  I'd prefer to avoid adding this argument to `getgauge`.

cjvogl:
While we could definitely get around using num_dim in getgauge to read the fort.gauges file, per your suggestion, I think there needs to be some way to tell the underlying GaugeData object that the gauges are 3D (visclaw/data.py:292).  In particular, the write() (amrclaw/data.py:226) and read() (amrclaw/data.py:253) of GaugeData currently rely on GaugeData.num_dim... which is set to a default value of 2 in init().

rjleveque:
I think we can avoid this by using the version in clawpack/amrclaw#128.

cjvogl:
That version in clawpack/amrclaw#128 looks good to me.  I'll remove all mentions of num_dim from this PR.

cjvogl:
This should be all set now.  My apologies for the confusion while I get over the learning curve for the structure of clawpack and using github PRs

rjleveque:
No worries, this is going as smoothly as any multi-repository set of changes and discussions could go...

PR #154, opened 2015-05-03
Deprecate `plot_topo_file` and `TopoPlotData`

mandli:


mandli:
So now nothing is shown unless someone is looking at the error logs.  I am not sure I like that so I may try to fiddle with the mechanism for emitting the warning.

rjleveque:
It does print the warning in an interactive session.  This seems fine to me.

PR #156, opened 2015-05-16
Fix frametools for 1d plotting

rjleveque:
@donnaaboise: One section of `frametools.py` assumed 2d and so merging #150 broke plotting in 1d.

This was for masking plots needed for kml files, so I moved it into a section marked
```
    if plotfigure.use_for_kml:
```
and indented.

Less complicated than it appears in the diff of files.


rjleveque:
Don't merge this yet -- this fix breaks other things.

Issue #155, opened 2015-05-16
ZIP64 needed for creating kmz file of large simulations


rjleveque:
I again ran into the problem discussed at the bottom of #150:

```
zipfile.LargeZipFile: Files count would require ZIP64 extensions
```

It's not the size of the files, it is the number that gives this error  -- in my case the `_plots` directory was only about 1GB but contained nearly 150,000 files, determined via
`find _plots -type f | wc -l`

Apparently only 65,535 are allow, e.g.
http://www.artpol-software.com/ziparchive/KB/0610051629.aspx

mandli:
This makes sense but what uses the zip format?

rjleveque:
Converting all the kml files into a single kmz file that can be opened in Google Earth.

donnaaboise:
I've enabled the ZIP64 extensions (see the googleearth branch). Does this fix the problem?  I don't have a large enough example to try.   

Also, you can use `zipinfo` to get the number of files in a zip file : 

    % zipinfo -t Tohoku_sift.kmz

should work.   The result differs slightly from what is reported by `find`, though, and I am not sure why.

donnaaboise:
Is this issue closed?

PR #157, opened 2015-05-19
Enable Zip64 extensions

rjleveque:
Allows creating kmz files for bigger simulations.


mandli:
@rjleveque did this fix the problem in #155?

PR #158, opened 2015-05-21
change patch.grid.c_edges to c_nodes in frametools.py 

rjleveque:
due to clawpack/pyclaw#517.

@mandli, @ketch:  VisClaw broke and I haven't tested everything, but I think this is the only change needed?

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -666,7 +666,7 @@ def plotitem2(framesoln, plotitem, current_data, stateno):
 
     # Grid mapping:
 
-    xc_edges, yc_edges = patch.grid.c_edges
+    xc_edges, yc_edges = patch.grid.c_nodes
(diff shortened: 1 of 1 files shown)


ketch:
After some grepping, I found that spot and no others, so I think you're right.

ketch:
Although it shouldn't be broken -- it should just give a deprecation warning.

ketch:
My mistake -- see https://github.com/clawpack/pyclaw/pull/518.  In any case, it's good that we're updating visclaw to the new names.  We need to update the docs too; I'll do that.  What about the other packages?

rjleveque:
I don't think any other packages would use this.

ketch:
Regarding the docs, it turns out we have no documentation on this outside of docstrings.  I will add some in the future, but we're fine as far as releasing 5.3.0.

PR #159, opened 2015-05-22
[WIP] Use output controller format rather than data object's

mandli:
I again ran into a case where the `plotdata` object's `format` was overridden somewhere and reset to ascii.  Using the `output_controller` instead and setting it's `file_format` value seems to be safer but this is really a non-intuitive interface since there already exists a `format` attribute in the data object.  Suggestions?


ketch:
The "right" solution seems to be for the `plotdata` object not to have a `format` attribute at all, but it's probably difficult to make that change.

mandli:
I agree.  I could see if there would be a way to extract all instances of the format to the `output_controller`.  As you said though this could be tricky and may break things unintentionally.

PR #165, opened 2015-09-17
Added amr_data_show attribute to suppress certain amr levels

cjvogl:
For backwards compatibility, included conditional that should default to showing all levels if amr_data_show is not set.


rjleveque:
Seems to work fine.  This is very useful for cases where one only wants to show coarser levels in making an animation quickly, for example.

Alternatively, I like that you can also set, e.g.
```
    plotitem.amr_data_show = [False, False, True]
```
to show only the fine patches.
level3

rjleveque:
@cjvogl: I just did a PR to your branch.  If you merge it in to your branch, it should then show up here...

cjvogl:
@rjleveque: all merged.

Issue #166, opened 2015-10-22
Fix Path to `gauge.data` file


mandli:
we have it in two places - the top directory, and the _output file. The plotting programs use the gauges.data in the top one, and the fort.gauges output from the _output dir. I suggest it be changed to use the _output only directory for both files, so that new runs with different gauges can be done at the same time as plotting. why do we even need it in the top dir?

mandli:
This was moved from clawpack/geoclaw#175 to here as the relevant code is in visclaw.

rjleveque:
All the .data files get created in the top dir when `make data` is executed and then are copied to `_output` when `make .output` is executed.  The plotting should take this data from the same place it's getting the output as suggested by @mjberger.  See PR #167 for a fix.

rjleveque:
Apparently fixed in #167.

PR #167, opened 2015-10-25
read gauge locations from outdir rather than datadir by default

rjleveque:


rjleveque:
This addresses #166.

mandli:
Is this the only spot where this is called?

rjleveque:
Oops, no.  It is also called in `plot_pages/plotclaw_driver`.  That could use a bit more work on the logic, I'll try to fix up and add to this PR.

PR #183, opened 2016-07-20
data.py should import GaugeSolution from PyClaw

rjleveque:
@mandli: can you check this is correct?

It was still working on my computer because `gauges.pyc` was still in my `amrclaw/src/python/amrclaw` directory but the `gauges.py` file was moved to `pyclaw` in the process of updating the gauge routines, I think.


mandli:
Yes, I have been meaning to track down how this bug got introduced and not caught in testing.  You can also remove the `GaugeData` import and line there as it is no longer used.

PR #185, opened 2016-07-31
Matlab : Fix use of 'break' when return should have been used

donnaaboise:
Minor change to compatibility with latest release of Matlab (15b)

### DIFF
--- a/src/matlab/setopengl.m
+++ b/src/matlab/setopengl.m
@@ -22,9 +22,9 @@ found_opengl = 0;
 for i = 1:length(rset),
   if (strcmp(lower(rset(i)),'opengl'))
     found_opengl = 1;
-    break;
-  end;
-end;
+    break
+  end
+end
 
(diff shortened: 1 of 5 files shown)


src/matlab/setopengl.m
```diff
@@ -22,9 +22,9 @@ function setopengl()
 for i = 1:length(rset),
   if (strcmp(lower(rset(i)),'opengl'))
     found_opengl = 1;
-    break;
-  end;
-end;
+    break
```
mandli: Isn't this still in the if statement?

donnaaboise: Yes, but it is nested in a for-loop, so it is okay, and the right thing to do, since I don't wan to return from the function, only break out of the for-loop.

mandli: Gotcha, was wondering if that was the case.

PR #187, opened 2016-08-14
remove plt.ioff() in frametools.py

### LINKED ISSUE #182: Interactive Plotting Only Plots First Figure
I ran into this when working with @mjberger yesterday and just saw this again on my own machine with a completely different example.  I am not sure when this started happening but my suspicion is that the parallel plotting may have caused something to break.

rjleveque:
This seems to be not needed and causes only first figure to appear.  

This seems to fix #182.


mandli:
Given that this has shown to fix #182 unless anyone has a use case that it broken by this I am going to merge this.

PR #188, opened 2016-08-16
Add a Python script to convert 2D clawpack data to VTK format 

xinshengqin:
Add a Python script to convert 2D clawpack data to VTK format, which can be visualized with Paraview.


mandli:
I know that the movie file is small but I am not sure if we want to start adding these to the repository.  I think it might make more sense in the eventual documentation that might be written for this feature though or alternatively provide a URL where the movie is kept.

xinshengqin:
I have removed the *.avi file.

rjleveque:
If you have the sample .avi file in another repository you might want to include a link to it in the README.md file.

dev/paraview/claw_to_VTK_2D/README.md
```diff
@@ -1,6 +1,9 @@
 # Convert Clawpack output data to VTK format
 
-The water\_depth.avi in this directory shows an example of visualizing result of
+An example, water\_depth.avi can be found at:
+https://github.com/xinshengqin/claw\_to\_vtk
```
rjleveque: This link doesn't seem to work as written, remove the back slashes?

xinshengqin: I am still testing it.

xinshengqin:
I think the link works now.

PR #190, opened 2016-12-01
Support Python 3

ketch:
See https://github.com/clawpack/clawpack/pull/97.


ketch:
Thanks for catching that.  Apparently tuple unpacking is a thing of the past.

On a separate note, it's a bit strange that we have that `examples.py` file in visclaw at all.  I guess JSAnimation was just copied wholesale into visclaw.

rjleveque:
Yes, it was copied wholesale and probably I should clean this up.

PR #196, opened 2017-01-28
Update changes to KML plotting capabilities

donnaaboise:
Updated code that reads in gauges.data file.  Anything after the actual gauge data (t1,t2,x1,x2) is ignored.

### DIFF
--- a/src/matlab/create_clines.m
+++ b/src/matlab/create_clines.m
@@ -30,11 +30,11 @@ while (1)
   if (mappedgrid == 1 | manifold == 1)
     if (mappedgrid == 1)
       if (nargin('mapc2p') == 2)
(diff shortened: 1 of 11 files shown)


mandli:
The code you wrote (python) looks ok but why not use the Gauges class instead so that if changes are made to the format you won't have to update again?

donnaaboise:
I guess I didn't know there was a gauges class.   Sounds like it would be much simpler.  We'll fix that and reissue the PR.   Thanks.

mandli:
It was added with the changes to the format so probably no reason you would have known.  Let me know if you need help using it.

donnaaboise:
I am closing this PR - Melody fixed the Gauges issue (which now uses the Gauges class) and so that will replace this one.   I'll issue a new one for the Matlab changes.

PR #198, opened 2017-02-04
modify plotpages.py to use gauge module for creating kml file

MelodyShih:
Using latest gauge class in kml plotting.


mandli:
@MelodyShih does this work now or are you waiting to merge this in with some of the other changes mentioned in #199?

MelodyShih:
@mandli yes, it works without the changes mentioned in #199.

PR #199, opened 2017-02-04
(matlab) Fix contour lines in multi-block case

donnaaboise:
Handle contour lines across multiple ForestClaw blocks


donnaaboise:
actually, hold off on this PR - I think I can implement it in a clearer way.

PR #202, opened 2017-02-10
Python 3 compatibility changes.

quantheory:
These two changes seem to be necessary when using Python 3.

Tested with Python 3.4.5 from an Anaconda installation.

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -1463,6 +1463,9 @@ def call_setplot(setplot, plotdata, verbose=True):
     Otherwise assume setplot is a function.
     """
     import types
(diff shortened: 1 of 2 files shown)


ketch:
The reload part will still fail for Python 3.0 to 3.3, it seems:

http://stackoverflow.com/questions/961162/reloading-module-giving-nameerror-name-reload-is-not-defined

It's probably worthwhile to make it work with those versions too, since it's so simple.

quantheory:
You caught me; I was being lazy and ignoring the old Python 3 versions. This last commit should fix those as well, though I don't actually have a machine with one of those versions on hand to test it.

ketch:
Thanks!

PR #205, opened 2017-02-15
Frametools imports

ketch:
This is intended to replace #204.  It moves almost all the imports in frametools to the top of the file, avoiding the reload issue discussed there.  One import (plotpages) is left in a function because putting it at the top throws an exception due to a circular import.  That circular import could be resolved in a better way by making `set_show()` a method of `ClawPlotData()`.  I didn't do that here since it is really a separate issue and deserves its own PR.

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -12,7 +12,6 @@ import numpy as np
 import re
 import logging
 import clawpack.clawutil.data as clawdata
-from . import gaugetools
 import time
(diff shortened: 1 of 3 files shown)


ketch:
I have tested this for PyClaw applications.  Please test it for other applications before merging.

mandli:
I just tested this for GeoClaw, AMRClaw, and Classic and everything seems to be working.

PR #208, opened 2017-02-21
Allow more options to be passed to ianimate(): figsize, xlim, ylim.

ketch:
One user was having trouble with the animation figures being too large on a small screen.  This allows the figure size and axis limits to be set in the ianimate() call, using the normal matplotlib keywords.

### DIFF
--- a/src/python/visclaw/ianimate.py
+++ b/src/python/visclaw/ianimate.py
@@ -1,9 +1,9 @@
 """
 Interactive animations in IPython notebooks using matplotlib and JSAnimation.
 """
-
-
 from __future__ import absolute_import
+
+
 def ianimate(frame_list,plotdata=None,ivar=0,varname=None,**kargs):
     """
         frame_list may be:
@@ -17,6 +17,10 @@ def ianimate(frame_list,plotdata=None,ivar=0,varname=None,**kargs):
     from clawpack.pyclaw import Controller
     import numpy as np
 
+    figsize = kargs.get('figsize', (10,6))
+    ylim = kargs.get('ylim', None)
+    xlim = kargs.get('xlim', None)
+
     if isinstance(frame_list,Controller):
         frame_list = frame_list.frames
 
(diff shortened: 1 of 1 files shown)


src/python/visclaw/ianimate.py
```diff
@@ -17,23 +17,34 @@ def ianimate(frame_list,plotdata=None,ivar=0,varname=None,**kargs):
...
+        if xlim is None:
+            ax.set_xlim(xc[0],xc[-1])
+        else:
+            ax.set_xlim(xlim)
+
+        if ylim is None:
+            ymax = max([np.max(f.q[ivar,:]) for f in frame_list])
+            ymin = min([np.min(f.q[ivar,:]) for f in frame_list])
+            ydiff = ymax-ymin
+            ax.set_ylim(ymin-ydiff/10.,ymax+ydiff/10.)
```
mandli: I probably just don't remember but why do we handle the x and y axes different?

ketch: I'm not sure which part you're referring to.  I wrote this code and it does what I usually want: tight axes in x, but a little extra space at the top and bottom in y (in case there is a constant state just at the top or bottom boundary).  Also, we know the x bounds immediately, but in order to avoid clipping data or auto-rescaling axes we have to look at all the frames for y.

mandli: Ah, good point.  I was thinking of 2D pcolor plot where the x and y directions should be treated the same.  Does this function know anything like that?

ketch: Wait...you mean there are problems that involve two space dimensions?   :-o

If you look at the file, you'll see that 2D images are handled further down, and for them x and y are handled in the same way.  However, your comment brings up the point that manually setting xlim or ylim only has an effect for 1D plots.  I guess it would be useful to let this be set also in 2D (e.g., to have an animation over a zoomed region).  I'll add that.

ketch: Okay, this works now.  Here's a demo:

https://gist.github.com/75607d3c9b8ae0fced9dd374751dd244

ketch: Ack, it didn't even save on image from the animation in the gist.  But if you download and run the notebook against this branch you'll see that it works just like one would hope.

mandli:
This works for me (the notebook).

PR #210, opened 2017-03-09
Remove all uses of hold() command.

ketch:
Our calls to `plt.hold(True)` now have no effect except to trigger
a deprecation warning.  This is because hold is set to True now by default.

**This PR is compatible only with matplotlib 2.0+**.

 It may be necessary in
some places to add calls to clf() since the default
behavior of matplotlib has changed (but that need is unaffected by this PR,
and that changed happened quite awhile ago, so I doubt that anything is needed).

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -179,7 +179,6 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
             axescmd = getattr(plotaxes,'axescmd','subplot(1,1,1)')
             axescmd = 'plotaxes._handle = plt.%s' % axescmd
             exec(axescmd)
-            plt.hold(True)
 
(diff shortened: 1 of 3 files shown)


ketch:
See http://matplotlib.org/api/api_changes.html#hold-functionality-deprecated

mandli:
Instead of removing these should we check the matplotlib version or did the change to `True` by default already occur?

ketch:
I believe it occurred a very long time ago. @rjleveque says it's been that way since he started using Python.

mandli:
I definitely do not remember although I stopped relying on `hold` awhile ago.  If the default behavior is fine back awhile I would say merge this as is.

ketch:
Yeah, my bolded statement about compatibility above was wrong.  It should have just said "the deprecation warnings that this avoids are only present in Matplotlib 2.0+".

PR #216, opened 2017-06-07
add plot_type 1d_pwconst option to frametools.

rjleveque:
Plot the solution as a piecewise constant on each grid cell, rather than with a piecewise linear function connecting cell averages at midpoints.

Useful for plotting 1d amr results in a way that better shows the different grid resolutions, e.g. the figure below for the Woodward-Colella blast problem (Ex. 15.1 in FVMHP), soon to be added to `amrclaw/examples`.
wcblast


ketch:
This is really nice.

PR #219, opened 2017-06-23
WIP: Properly set pobj and don't make colorbar if this is None

rjleveque:
I was seeing strange behavior in GeoClaw plots in Python3, that sometimes the colorbar showed up all green (from the land plots) rather than showing the colorbar for the ocean plot, even though that was the only one requested.

Also when plotting data that was entirely masked on all patches, the gauge was plotted where the colorbar should have been.

I figured out there was a problem in how `exec` was used that meant `pobj` wasn't getting set at all. Moreover `exec` and `colorbar` both work differently in Python3.

I've rewritten the way `pobj` is set when calling the `pcolor` function for this particular `plot_type`, to put all the arguments into `kwargs` (rather than constructing a string `pcolor_cmd` and then using `exec`).  

I suggest we might want to do the same for all other `plot_types` and I'll work on that unless anyone has a better idea how to do this.

I also modified the code so that if `pobj` ends up as `None`, meaning all patches were masked, then it does not make a colorbar.  But then the main plot ends up bigger than it would otherwise, unpleasant in an animation with other frames.  I think we might want to rewrite all the colorbar stuff to use `colorbar.make_axes_gridspec` to always make the colorbar axes, and probably we should make it easier for the user to specify other attributes of colorbars (e.g. location, orientation).

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -203,6 +203,11 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
                 # loop over patches:
                 # ----------------
 
+                for itemname in plotaxes._itemnames:
+                    # initialize before loop in case all patches are masked:
+                    plotitem = plotaxes.plotitem_dict[itemname]
+                    plotitem._current_pobj = None
+
                 for stateno,state in enumerate(framesoln.states):
 
                     patch = state.patch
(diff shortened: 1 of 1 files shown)


mandli:
I have been seeing something similar lately as well when playing with some methods for reducing the memory footprint when plotting 1,000s of patches (colorbar and axes are screwed up).  I am not sure if these are related but I like the idea of moving towards removing `exec` commands as much as possible.

rjleveque:
This is a big enough job that I think we should leave it for the next release, and put a note in that plotting with Python3 may have some problems currently.

mandli:
Another way to do this that keeps some of the `kwargs` separate and maintains the ability to do a particular `pc_cmd` (for `pcolor`):
```python
        vmin = pp['pcolor_cmin']
        if vmin is 'auto':
            vmin = None
        vmax = pp['pcolor_cmax']
        if vmax is 'auto':
            vmax = None

        edgecolors = None
        if pp['celledges_show']:
            edgescolors = pp['celledges_color']

        if not var_all_masked:
            plot_cmd = getattr(plt.gca(), pc_cmd)
            pobj = plot_cmd(X_edge, Y_edge, var, cmap=pp['pcolor_cmap'],
                            vmin=vmin, vmax=vmax, edgecolors=edgecolors,
                            **pp['kwargs'])
```
This was an attempt at fixing the problem in clawpack/geoclaw#275 which did not work.

mandli:
I ran into something similar in Python 3 again.  I changed assignment of `pobj` to `plotitem._current_pobj` so that it gets set by the last non-masked plot object.  This could be a quick fix for the rest of the `plotitem`s that are having this issue.

rjleveque:
Does this also work for making a colorbar when all patches are completely masked?  This happens in  what we're now testing with using the adjoint to flag for refinement and want to plot the flagged cells.  At later times it may happen that no cells are flagged and all patches are masked.  I suspect this has to be dealt with differently since it doesn't know the colormap to use in the colorbar if there's no pobj the colorbar goes with.

PR #222, opened 2017-08-28
Auto detect file format, second attempt

rjleveque:
This requires clawpack/pyclaw#581.

The default `ClawPlotData.format` value is now `None`, indicating that pyclaw should figure it out for itself.

I added a function `ClawPlotData.getframe_time` that is now used in `plotpages.py` to only grab the times, in place of the deprecated `OutputController` function.

While I was at it, I also added a function `ClawPlotData.getframe_list` that returns a list of all the `framenos` found in `plotdata.outdir` and the corresponding time, and a function to `Iplotclaw` so that this can be easily invoked at the interactive `PLOTCLAW>` prompt, with output like this:

```
PLOTCLAW > list
Frames found in _output:
Frame number 0 at time t = 0
Frame number 1 at time t = 0.25
Frame number 2 at time t = 0.5
Frame number 3 at time t = 0.75
Frame number 4 at time t = 1
```

I've been meaning to do this for a long time!

### DIFF
--- a/src/python/visclaw/Iplotclaw.py
+++ b/src/python/visclaw/Iplotclaw.py
@@ -235,6 +235,7 @@ class Iplotclaw(Iplot):
         print('plotgauge    : loop through plots of all gauges')
 
     def do_list(self, rest):
+        print('Frames found in %s:' % self.plotdata.outdir)
         frame_list = self.plotdata.getframe_list(self.plotdata.outdir)
         for frameno_t in frame_list:
             print('Frame number %i at time t = %g' % frameno_t)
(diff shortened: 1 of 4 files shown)


mandli:
Looks good to me.  I really like the listing!

ketch:
There is actually a `read_t()` function in the pyclaw/fileio/*.py files; it might be nice to use that here so that we don't need to maintain two functions that do the same thing.   Should make this code simpler, too.  And `petclaw/fileio/petsc.py` has a `read_t()` function too, so this should also then work for those files.

PR #227, opened 2018-05-14
add ClawPlotItem.colorbar_kwargs

rjleveque:
useful e.g. for adding arrows to colorbar when colors saturate, e.g.
    plotitem.colorbar_kwargs = {'extend': 'both'}


mandli:
I at some point did the same thing but only locally.  Maybe we should also just add the other too the dict as well?

rjleveque:
You mean add things like `colorbar_shrink`?  We could eliminate such things, but for backward compatibility I think we can leave them as attributes and the user could set them either explicitly or as part of `colorbar_kwargs`.

rjleveque:
@mandli, is this ok now?

Issue #223, opened 2018-02-21
Use JSAnimation directly from matplotlib


ketch:
I discovered today that JSAnimation was merged into matplotlib version 2.1, last fall.  We should switch over to using this.  As far as I can tell, they fixed the issue where you would sometimes get a javascript error instead of the animation (hooray!)  You will need to update to matplotlib 2.1+ for this to work.  There is a modified version of animation_tools.py in clawpack/riemann_book (currently in a PR) that can be backported to visclaw (and we can delete the JSAnimation directory).

mandli:
I am glad this capability has been merged in and we don't need to keep our own version (which is behind the newest version as well).

ketch:
Any opinions on when to make this change?  It will break things for people using Matplotlib < 2.1.0, which was released in October 2017.  I think it would be nice to include this change in the next release of Clawpack, which I guess is coming very soon.

mandli:
+1.

ketch:
If you have a branch and want to push it and mention what's not working, I can also take a look.

rjleveque:
Closing this issue for now since we ran into problems using JSAnimation from mpl.  Something to consider again in the future perhaps..

PR #228, opened 2018-05-28
WIP:  use JSAnimation from mpl if available

rjleveque:
Addresses #223. 

Try to use JSAnimation from mpl if available, if not try to use local JSAnimation.

Leaving the local version in for now for two reasons:
1. Many users may not have an up-to-date matplotlib,
2. The mpl version is coming out ugly at the moment relative to old version.

Presumably the second issue can be fixed by improving the resolution of what is displayed by `imshow`.   Compare:

 - [local JSAnimation](http://faculty.washington.edu/rjl/misc/jsanimate_tests/_plots_local/movieframe_allframesfig2.html)
 - [from matplotlib](http://faculty.washington.edu/rjl/misc/jsanimate_tests/_plots_mpl/movieframe_allframesfig2.html)

### DIFF
--- a/src/python/visclaw/animation_tools.py
+++ b/src/python/visclaw/animation_tools.py
@@ -2,9 +2,9 @@
 This animation_tools module contains tools to create animations in Python and
 Jupyter notebooks.
 
-Three types of animations are supported: 
- - using the ipywidget interact to create a figure with a slider bar, 
- - using JSAnimation to create Javascript code that loops over a set of 
+Three types of animations are supported:
+ - using the ipywidget interact to create a figure with a slider bar,
+ - using JSAnimation to create Javascript code that loops over a set of
    images and adds controls to play as an animation.
  - creation of mp4 files using ffmpeg (provided this package is installed).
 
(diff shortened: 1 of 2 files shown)


ketch:
I never liked the approach of writing to file and then loading back to create the animation, but given the existing visclaw infrastructure it seems difficult to avoid.  In any case, I don't understand what part of the process is different between using our JSAnimation and the one in matplotlib -- they both use `imshow`, right?

I think users will not understand the warning message; perhaps it's better to have it say something like  "Warning: in the future, creation of animations will require Matplotlib 2.1 or greater.  Please update your matplotlib."

ketch:
Sorry, I was mixed up about the warning messages -- they actually make a lot of sense as they are.

One thing that isn't very important: I believe jsanimation was incorporated in matplotlib 2.1.0:

https://matplotlib.org/users/prev_whats_new/whats_new_2.1.0.html#interactive-js-widgets-for-animation

rjleveque:
Yes, it would be nicer to make the animation directly from a list of fig's rather than png's. This would require some refactoring but probably worth doing eventually.

I can't figure out why `to_jshtml` gives different results than the old JSAnimation, maybe I can get @jakevdp to help me out with this at some point.

rjleveque:
Various dpi and figure size defaults [changed in matplotlib 2.0](https://matplotlib.org/users/dflt_style_changes.html) that seem to affect how well a png read in as an image gets rendered.  Experimenting with figsize and dpi on figures and savefigs, I have not been able to reproduce as nice resolution in animations as we used to see, at least not without increasing the figure sizes substantially to where they often don't fit in the browser window nicely.  

One can get something close by doing
```
import matplotlib as mpl
mpl.style.use('classic')
```
but it's still not great.

rjleveque:
Closing this for now.  Additional testing has not helped resolve the issues with the JSAnimation now built into matplotlib, so let's stick with the local version for now.

PR #229, opened 2018-06-05
Remove all uses of hold() command.

ketch:
This command now has no effect except to trigger
a deprecation warning.  It may be necessary in
some places to add calls to clf() since the default
behavior of matplotlib has changed.


rjleveque:
I think this is out of date -- these changes were merged in #210 already.

mandli:
I agree with @rjleveque, this does seem to be already merged.  It looks like @ketch's master branch is behind but I am not sure why the PR isn't comparing correctly.

ketch:
Right, thanks. I was cleaning out old branches and got mixed up.

PR #230, opened 2018-06-12
Change to new structure for current pobj

mandli:
This fixes the recurrent land colorbar problem that pops up every once in awhile.  It has to do with the semi-random last object that is used to form the colorbar.  I imagine that this could be fixed in another way but this is simple and seems to work.

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -738,7 +738,7 @@ def plotitem2(framesoln, plotitem, current_data, stateno):
 
     if pp['plot_type'] == '2d_pcolor':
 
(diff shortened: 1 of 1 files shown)


rjleveque:
The comment in line 979 should then be updated to
```
    try:
        plotitem._current_pobj = pobj
    except NameError:
        pass # if no pobj generated
```

But eventually we should make it uniform and only set `plotitem._current_pobj` everywhere in this code, removing the local `pobj`?

mandli:
Fixed.  I think moving to `_current_pobj` overall would be better but doing it with a search and replace was scary at the time.

PR #233, opened 2018-08-14
Minor updates to Matlab and KML plotting

donnaaboise:
* Add a default DPI to tikz plots (to get vector graphics for AMR plots)
* Fixed problem with gauges not showing up by default in Google Earth plots.


rjleveque:
Thanks!

PR #235, opened 2018-12-26
Binary reader for Matlab

donnaaboise:
Added reader for binary output for AMRClaw.  

* 2d code assumes that ghost cells are included in output and must be stripped.
* 3d assumes ghost cells not written to binary output

* Several other changes to formatting and warnings were fixed - so lots of changes that are not significant.


rjleveque:
Merging this, but at some point we should resume the discussion of how to allow variable number of ghost cells in both Fortran output and Python reading, for both ascii and binary.  See clawpack/amrclaw#236

PR #237, opened 2019-02-09
improve time axis on gauge plots

rjleveque:
add attribute  for label and  to rescale t, e.g. from seconds to hours by setting time_scale = 3600.

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -667,6 +667,9 @@ class ClawPlotAxes(clawdata.ClawData):
         self._next_ITEM = 0
         self.add_attribute('figno', self._plotfigure.figno)
 
+        self.add_attribute('time_label', 'time')  # for time axis in gauges
+        self.add_attribute('time_scale', 1)  # to rescale time (e.g. to hours)
+
     def new_plotitem(self, name=None, plot_type=None):
         # Create a new entry in self.plotitem_dict
 
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -667,6 +667,7 @@ class ClawPlotAxes(clawdata.ClawData):
         self._next_ITEM = 0
         self.add_attribute('figno', self._plotfigure.figno)
 
+        # attributes for gauge plots
         self.add_attribute('time_label', 'time')  # for time axis in gauges
         self.add_attribute('time_label_kwargs', {})  # kwargs for xlabel cmd
         self.add_attribute('time_scale', 1)  # multiplicative factor to rescale t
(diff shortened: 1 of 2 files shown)


mjberger:
Can we get units in to the plots somehow?

rjleveque:
With this PR you can do, e.g.

```
    plotaxes.time_scale = 3600.  # convert to hours
    plotaxes.time_label = 'time (hours)'
```
so the label has the proper units.  The default is just 'time' with no units since we don't know what units the user's code is using.  (It's seconds by default for GeoClaw, but not necessarily for other Clawpack applications.)  

You can also set 
```
    plotaxes.time_label = None
```
to recover the current default behavior of no label on the horizontal axis.

One thing that occurs to me now is that there should also be a `time_label_kwargs` parameter to allow specifying arguments to `label`, e.g. if the user wants a bigger fontsize for the label.  I'll add this to this PR.

When I get time I plan to add similar attributes to `frametools.py` to allow setting `xlabel` and `ylabel` without needing an `afteraxes` function.  

I thought about also allowing rescaling of `x` and `y` (e.g. from meters to km would be useful in some of our codes), but I don't think so, since this can already easily be done by providing a `mapc2p` function and allowing this additional mechanism may lead to confusing conflicts with that.  Opinions?

mandli:
We currently often do this for plotting storm surge.  It would be interesting to try and change over to using this rather than our custom `afteraxes`.

mandli:
I cannot seem to get this to work.  I removed my `afteraxes` function and added the attributes above but the plots turn out empty.  This is what I have in my `setplot.py`:
```python
    plotfigure = plotdata.new_plotfigure(name='Gauge Surfaces', figno=300,
                                         type='each_gauge')
    plotfigure.show = True
    plotfigure.clf_each_gauge = True

    # Set up for axes in this figure:
    plotaxes = plotfigure.new_plotaxes()
    # plotaxes.xlimits = [-2, 1]
    plotaxes.time_scale = 3600.0
    plotaxes.time_label = 'time (hours)'
    plotaxes.title = 'Surface'

    # Plot surface as blue curve:
    plotitem = plotaxes.new_plotitem(plot_type='1d_plot')
    plotitem.plot_var = 3
    plotitem.plotstyle = 'b-'
```

rjleveque:
Try changing it to:
```
    plotaxes.time_scale = 1./3600.0
```
Sorry, I changed to a multiplicative factor in https://github.com/clawpack/visclaw/pull/237/commits/3225e3b13134dc1e4bd692b96c502e0fce7af4f6

Not sure if that's the problem unless you also have `xlimits` set to something that makes it not show up?

mandli:
Hmm, that did not seem to work and I do not have a `xlimits` set.  It is plotting from `[0.0, 1.0]` on the x-axis though.  Not sure where else to look.

rjleveque:
Seems to work ok now for @mandli.

PR #239, opened 2019-02-24
Add animation size and `bbox_inches='tight'` options

mandli:


rjleveque:
At some point we should also allow `dpi` to be a parameter to `savefig` and/or to `figure` since wider figures need more resolution.  It's not clear where best to do this, and it will take a bit of refactoring to expose these options to the user in `setplot.py`.  

But for now this is already useful so I will merge it.

Issue #244, opened 2019-06-26
matplotlib version 3 issue


cheginit:
I have installed clawpack from git repo with matplotlib version 3 and python 3.7. When I try to run any example it fails at the animate step with the following error:
```bash
2019-06-25 08:29:06,839 INFO CLAW: Animation.save using <class 'clawpack.visclaw.plotpages.plotclaw_driver.<locals>.myHTMLWriter'>
Traceback (most recent call last):
  File "/home/taher/repositories/github/clawpack/visclaw/src/python/visclaw/plotclaw.py", line 166, in <module>
    plotclaw(sys.argv[1], sys.argv[2], sys.argv[3])
  File "/home/taher/repositories/github/clawpack/visclaw/src/python/visclaw/plotclaw.py", line 152, in plotclaw
    plotpages.plotclaw_driver(plotdata, verbose=False, format=format)
  File "/home/taher/repositories/github/clawpack/visclaw/src/python/visclaw/plotpages.py", line 2744, in wrapper
    return f(*args, **kwds)
  File "/home/taher/repositories/github/clawpack/visclaw/src/python/visclaw/plotpages.py", line 3079, in plotclaw_driver
    anim.save(fname, writer=myHTMLwriter)
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 1156, in save
    writer.grab_frame(**savefig_kwargs)
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 232, in saving
    self.finish()
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 528, in finish
    MovieWriter.finish(self)  # Will call clean-up
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 367, in finish
    self.cleanup()
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 531, in cleanup
    MovieWriter.cleanup(self)
  File "/home/taher/applications/anaconda/envs/claw/lib/python3.7/site-packages/matplotlib/animation.py", line 399, in cleanup
    out = TextIOWrapper(BytesIO(out)).read()
[...]

PR #245, opened 2019-06-26
WIP: use anim.to_jshtml if available

rjleveque:
This address #244, but not yet optimized.  We ran into this before when trying to switch to `anim.to_jshtml` to make JS Animations: for some reason the resolution of the resulting animation is much worse than that of the png files being read in.  Comparison of old and new below.

Maybe someone knows how to fix this?

old
new

### DIFF
--- a/src/python/visclaw/plotpages.py
+++ b/src/python/visclaw/plotpages.py
@@ -3034,7 +3034,8 @@ def plotclaw_driver(plotdata, verbose=False, format='ascii'):
             fname = '*fig' + str(figno) + '.png'
             filenames=sorted(glob.glob(fname))
             fig = plt.figure()
-            im = plt.imshow(Image.imread(filenames[0]))
+            # using 'nearest' gives slightly better resolution:
+            im = plt.imshow(Image.imread(filenames[0]), interpolation='nearest')
             plt.axis('off')  # suppress second axis around image
             def init():
                 im.set_data(Image.imread(filenames[0]))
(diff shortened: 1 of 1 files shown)


ketch:
One solution is to avoid writing figure files altogether.  I think this will be the most robust long-term, but I realize it may be significant work to implement.

rjleveque:
I swapped the order or the try/except blocks so it first tries our original approach of using JSAnimation directly and only reverts to anim.to_jshtml if that fails, since we still haven't figured out how to make that give movies with full resolution of the png files.

Several users have run into this bug with the code in master recently and I have to tell them to merge in this PR, so I suggest we go ahead and merge it, so their code will at least produce ugly animations rather than dying.

I did try again to figure out what's going on.  The problem is not the resolution of the png files being saved since opening them alone looks fine, and so does the JSAnimation version of the animation.  Somehow it is in `imshow`.  I discovered there's a [figimage](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.figimage.html) command that maybe does what we want better, but I couldn't get it to work.

@ketch suggested avoiding png's altogether, which for some applications would work. But it is very useful to also have the png of every frame separately and the other webpages produced with `plotpages.py` allow looking at these separately, so we need to generate them anyway.  And for some applications (e.g. many tsunami models with huge grids and several figures for each of many frames) it can take hours to make the plots, so plotting two versions of every figure isn't viable, nor could we save them all in memory to animate directly.

donnaaboise:
Thanks, Randy.  I am still a bit confused as to the source of the error, because I can create plots that look fine to me, and the movies work as well.  We just had problems on the Windows install.  But maybe I am not up-to-date with the latest Anaconda?

rjleveque:
It depends a lot on what particular version of matplotlib you have.  The error you sent me by email,

```
 out = TextIOWrapper(BytesIO(out)).read()                                                               
TypeError: a bytes-like object is required, not 'str'                                    
```

is the same one encountered in https://github.com/clawpack/visclaw/issues/244, I believe, which comes up with newer versions of matplotlib when trying to use the older JSAnimation code.  On the other hand sufficiently old versions of matplotlib don't have the `to_jshtml` function.

PR #251, opened 2019-10-23
fix mapc2p functionality for plotitem1

cr2940:
In Frametools, there is a bug which was found by trying an example with 1D advection when there is a mapc2p function in setplot. Namely, the `current_data.x` is just a scalar, and the mapc2p function acts only on a scalar, giving back only a scalar for `p_centers`. But using `c_centers` fixes this.


mandli:
This looks good to me.

PR #255, opened 2020-02-11
new gridtools.py module for dealing with AMR output

rjleveque:
Can be used to map data from collection of patches to a
2D uniform grid or to a 1D transect, for example.

For more details and examples, see
http://www.clawpack.org/new_features_for_v5.7.0/notebooks_html/gridtools.html


mandli:
Love this!  Do you want to setup any tests or just use the notebook to test?  The latter you can just extract the Python code and make sure it runs to completion.

PR #260, opened 2020-04-10
Animation improvements

rjleveque:
I think I finally got the `anim.to_jshtml` working better so the quality of animations on html pages created by `make plots`  is much better than what's shown in #245.  I'm not sure why, since I am now calling a function from `animation_tools` that does (I think) essentially exactly the same thing as the earlier code, but somehow it works better.  I still don't completely understand what's going on.

Anyway, I have now removed the dependence on our old `JSAnimation` completely and just use what's built in to matplotlib.  So this won't work with old versions of matplotlib, but since we are now only supporting Python3 officially I think this is ok?

Others should try it out and let me know about any problems before merging.

This PR also includes several changes to `animation_tools` to fix some things that were broke and to give greater functionality.  I've also been testing this on some notebooks, in particular some in the apps repository and I'll do a PR for those.

This could still use some cleaning up, at some point.

### DIFF
--- a/src/python/visclaw/animation_tools.py
+++ b/src/python/visclaw/animation_tools.py
@@ -4,8 +4,10 @@ Jupyter notebooks.
 
 Three types of animations are supported: 
  - using the ipywidget interact to create a figure with a slider bar, 
- - using JSAnimation to create Javascript code that loops over a set of 
-   images and adds controls to play as an animation.
+ - an `animation.FuncAnimation` object, which can be displayed using the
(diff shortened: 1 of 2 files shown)


ketch:
Thanks for revisiting this -- it will be really nice to rely directly matplotlib for this.  I don't have time to do testing right now but I did look quickly through the code and it makes sense to me.

mandli:
Works fine for me, thanks @rjleveque!

Issue #246, opened 2019-08-16
Create One Figure and Change `xlimits` and `ylimits` to Reduce Computation


mandli:
Basic idea is to modify the plotting loop so that only one figure for a given field is created and then if zoom ins are requested that instead of creating a figure for each that we just adjust the `xlimits` and `ylimits` accordingly.

PR #262, opened 2020-04-11
Skip patches not seen

rjleveque:
When looping over patches this version skips a patch if it has no intersection with the region specified by `plotaxes.xlimits` and `plotaxes.ylimits`.  This speeds up plotting substantially for some problems, if you want to zoom in on different regions in different plots and have thousands of patches that can be ignored in each case.   This is distinct from what's proposed in #246 but useful in some contexts.

Note a possible problem is that if a user creates a figure with specified `xlimits` and `ylimits` but then gives additional `xlim`, `ylim`, or `axis` commands to zoom out, then some patches will not be seen that should be seen in the wider view, which could be misleading or confusing.  I'm ok with this but others may differ.

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -209,6 +209,23 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
                 for stateno,state in enumerate(framesoln.states):
 
                     patch = state.patch
+                    if (plotaxes.xlimits is not None) \
+                            & (type(plotaxes.xlimits) is not str):
+                        if (patch.dimensions[0].lower \
+                                    >= plotaxes.xlimits[1]) \
+                                or  (patch.dimensions[0].upper \
+                                    <= plotaxes.xlimits[0]):
+                            continue  # go to next patch
+
+                    if len(patch.dimensions) > 1:
+                        # 2d patch
+                        if (plotaxes.ylimits is not None) \
+                                & (type(plotaxes.ylimits) is not str):
+                            if (patch.dimensions[1].lower \
+                                        >= plotaxes.ylimits[1]) \
+                                    or  (patch.dimensions[1].upper \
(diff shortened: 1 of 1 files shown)


mandli:
This is great!

mjberger:
So if you zoom out, what do you have to do for “real” re-plotting that includes the missing patches?   

Does is also skip patches for batch?

mandli:
I am not sure this would make sense for interactive plotting given that a user could move the view into the plot but the `xlimits` and `ylimits` would not change.  That may be something to add (check for interactive plotting).  This should work fine for batch plotting.

rjleveque:
Yes, it should work fine for batch processing, also in parallel.

When zooming out on a plot there's no easy way to recognize that and now plot the missing patches.  So the use case I had in mind was where the appropriate `xlimits` and `ylimits` are set in `setplot.py` for each `PlotAxes` object, and if you want multiple views at different zooms the user would have to set up multiple plot figures in `setplot.py`.  And note that if you are doing batch processing, i.e. using `make plots` to make the html pages rather than `Iplotclaw` for interactive plotting, then there is no way to zoom out other than specifying another figure, so this approach is ideal for batch processing to avoid spending time plotting patches not visible.  (Though what is suggested in #246 could be even faster in some contexts.)

For exploratory interactive work using `Iplotclaw` it is sometimes convenient to be able to zoom out and see all the patches.  So maybe this should be modified to somehow allow the user to specify this.  For example, we could add a new attribute to `ClawPlotAxes` so that the user could specify (in `setplot`),

        plotaxes.plot_all_patches = True

to revert to the old behavior, whereas by default this attribute would be `False` and so existing `setplot` files would lead to the new behavior.   Or maybe a better name would be `plot_patches_outside_extent` or something else.

What do you think?

mjberger:
I think that \'s good idea (a new plot-all-patches variables to remind user of the new behavior as well as how to revert).

PR #264, opened 2020-04-15
Respect plotdata.html_index_fname 

rjleveque:
rather than hardwiring _PlotIndex.html on movie page


mjberger:
what is the motivation for this?

rjleveque:
The user can set this in `setplot.py` and it was already respected everywhere but in the movie pages.

Usually the default `_PlotIndex.html` works fine but I discovered when pushing examples to github for documentation pages that links starting with underscore don't always work.

PR #268, opened 2020-05-10
close fig by default in printfig

rjleveque:
To avoid running out of memory when making lots of big plots
Can call with close_fig=False to suppress this


mandli:
Does this also avoid the "too many figures open" warning?

rjleveque:
Yes, it should.  Also, on the CSDMS HPC cluster it was running out of memory for some huge plotting jobs.

PR #270, opened 2020-06-03
Revert to BoundaryNorm if TwoSlopeNorm not found

rjleveque:
In add_colormaps, the improved version doesn't work with older versions of
matplotlib.

### DIFF
--- a/src/python/visclaw/colormaps.py
+++ b/src/python/visclaw/colormaps.py
@@ -152,7 +152,7 @@ def add_colormaps(colormaps, data_limits=[0.0,1.0], data_break=0.5,
     try:
         # Use this norm for proportional pw linear with better tick mark locations:
         norm = colors.TwoSlopeNorm(data_break, data_limits[0], data_limits[1])
-    except:
+    except AttributeError:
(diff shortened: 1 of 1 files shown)


src/python/visclaw/colormaps.py
```diff
@@ -149,8 +149,12 @@ def add_colormaps(colormaps, data_limits=[0.0,1.0], data_break=0.5,
...
     #norm = colors.BoundaryNorm(boundaries=bounds, ncolors=N)
 
-    # Use this norm for proportional pw linear with better tick mark locations:
-    norm = colors.TwoSlopeNorm(data_break, data_limits[0], data_limits[1])
+    try:
+        # Use this norm for proportional pw linear with better tick mark locations:
+        norm = colors.TwoSlopeNorm(data_break, data_limits[0], data_limits[1])
+    except:
+        # older matplotlib.colors did not have TwoSlopeNorm, revert to old:
+        norm = colors.BoundaryNorm(boundaries=bounds, ncolors=N)
```
mandli, line 157: Might be good to catch only the specific error, whatever that was.

rjleveque: Fixed.

PR #274, opened 2020-06-17
Added utility routines for geoclaw plots

donnaaboise:
Add three routines that read gauge data, plot gauges on 2d plots, and add time dependent regions.


rjleveque:
@donnaaboise: I don't know if anyone else is using matlab plotting now to test this out, but if you think it's good to go we can merge it in.

donnaaboise:
Can we merge this in?  I just added comments to add_gauges and add_regions.  

I still find Matlab useful, especially for students who sometimes have trouble navigating the visclaw commands in set plots.py

rjleveque:
Thanks for adding these!

Issue #278, opened 2020-09-21
skip_patches_outside_xylimits doesn't work with grid mappings


rjleveque:
The new feature added in #262 skips plotting AMR patches that are outside the specified xlimits and ylimits, unless `plotaxes.skip_patches_outside_xylimits == False` as provided in #263.   By default this is set to `True`.

I recently discovered that this doesn't work well with a mapped grid where a `mapc2p` function is provided to map the computational `xc, yc` to the physical space `xp, yp` for plotting.  The problem is that `xlimits, ylimits` are specified in physical coordinates (what you see in the plot) whereas the check on whether to skip a patch is done before `mapc2p` is applied, and hence the `xc ,yc` used in the test are in computational coordinates.  A patch may easily appear to be outside the limits even though after the mapping it will be inside.  This is also a problem in 1d if `mapc2p` is specified to map `xc` to `xp`.

This would be fairly easy to fix if `mapc2p` were also a `ClawPlotAxes` attribute, and it seems like it probably should be since the axes should presumably be in the same coordinate system regardless of what item is being plotted on it.

However, for some reason (my bad design, probably), `mapc2p` is an attribute of a `ClawPlotItem`, or if that is not set, the code also checks for `mapc2p` as an attribute of `plotdata`, but there is no `plotaxes.mapc2p`.

Cleaning this up will take some work and needs further discussion.

In the meantime, users should be aware that if using a `mapc2p` function, you should also set `plotaxes.skip_patches_outside_xylimits` to `False`.  

Should this be the default, rather than `True`?

mjberger:
I agre e- False should be the default.  It would be an impossible one for users to debug. But
those in the know will know to set the flag to True.


 — Marsha

mandli:
I would favor keeping the default as `True` myself.

The reasoning is that there are very few who use mapped grids to plot and I would consider those who do to be the experts.  I also think that the speedup that this enables for the generic GeoClaw user, who is more than likely not an expert, would be well worth the issue as @rjleveque outlined it.  Furthermore I think that adding `mapc2p` to the plotting capabilities is something we should enforce as was done in the old Matlab plotting code.

Of course just my 2¢.

mjberger:
I used mapped grids, and I would not remember to do that. Not an expert!

 — Marsha

mandli:
@mjberger oh man, if you are not an expert are any of us except for maybe @rjleveque 😉

rjleveque:
[...]

PR #279, opened 2020-09-22
Don't skip patches if mapped grid used

rjleveque:
To address #278.

If the user hasn't set `plotaxes.skip_patches_outside_xylimits` then the internal `skip_patches_outside_xylimits` will be set to `True` only if there is no `map2cp` function specified for the `plotdata` or any `plotitem`.  

We need a better approach in general.

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -187,8 +187,7 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
             current_data = run_str_or_func(beforeaxes,current_data)
                 
             skip_patches_outside_xylimits = plotaxes.skip_patches_outside_xylimits
-            print('+++ skip_patches_outside_xylimits = ', \
-                  skip_patches_outside_xylimits)
+            
             if skip_patches_outside_xylimits is None:
                 # User didn't set.  Set to True unless there's a mapped grid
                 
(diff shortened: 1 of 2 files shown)


src/python/visclaw/frametools.py
```diff
@@ -186,6 +186,21 @@ def plot_frame(framesolns,plotdata,frameno=0,verbose=False):
             beforeaxes = getattr(plotaxes,'beforeaxes',None)
             current_data = run_str_or_func(beforeaxes,current_data)
                 
+            skip_patches_outside_xylimits = plotaxes.skip_patches_outside_xylimits
+            
+            if skip_patches_outside_xylimits is None:
+                # User didn't set.  Set to True unless there's a mapped grid
+                
+                MappedGridExists = (plotdata.mapc2p is not None)
```
mandli, line 194: Just a minor point for Python style.  Variables that are not classes are usually lower case and underscored so that `MappedGridExists` should be `mapped_grid_exists`.

rjleveque: Good point, I changed it to `mapc2p_exists` as more descriptive.

I chose the original name because there is also a `plotitem` attribute named `MappedGrid`.   But I don't check that in the test done in this PR since it seemed simpler and more cautious to just look for a `mapc2p` attribute even if `MappedGrid` is `False` (An option allowed in case the user wants to see things in the computational plane even if there is a grid mapping.)

rjleveque:
Oops, @mandli merged before I made another change of name.  I'll open a new PR.

src/python/visclaw/frametools.py
rjleveque: See name change in #280.

mandli:
Ah sorry!

PR #282, opened 2021-08-06
2D Tiff plot option

AsianHam:
Created an plotting option that plots via GDAL and creates TIFF files

To use, set plot type to '2d_gtiff'

Currently does not work on macOS.


donnaaboise:
How are you installing GDAL? I've had some trouble installing it recently - I use it for supporting plot2kml (for Google Earth plotting) in VisClaw.

mandli:
Heh, that has been a source of frustration for awhile.  GDAL versioning also is an issue.

rjleveque:
I've been using conda environments and haven't had problems with gdal recently, although I don't think I'm using many parts of it, so there may be issues I'm not aware of.

The .yml file I use includes (among many other things):

```
channels:
- conda-forge
dependencies:
- gdal
- libgdal
```

rjleveque:
@AsianHam:  Thanks for contributing a PR!

Unfortunately, the `gdal_merge.py` file that is included in this PR is GPL licensed, and hence we cannot include it in Clawpack (without making all of Clawpack GPL, which we don't want to do).

Since that code comes from a different developer, I think it would be best to simply list it as a dependency, with a pointer to where to obtain the code for those who want to use it.  And with an informative error message for users who don't have it installed.

Also, I'd like to try this out, but am working on a Mac. What is the issue that keeps it from working on MacOS?

mandli:
I have not any issues installing GDAL but the significant version differences have been an issue.  The `conda` version for instance I believe is pretty old at this point, which also causes problems.  The major issue we did have though *we think* was a weird incompatibility with `libtiff` library and GDAL that is used with the GeoTIFF functionality.  This seemed to be only a problem on MacOS but it was a versioning problem as far as I could see (version incompatibilities between GDAL and `libtiff`).

AsianHam:
I'll remove gdal_merge and pr again.

macOS gives "unknown attribute" errors when gdal attempts to edit the
metadata in the TIFF files.

I've installed gdal via brew and python (pip install), both seem to have
the same results on macOS.

AsianHam (commented):
removed gdal_merge

mandli:
Do you need `gdal_merge` though to do AMR plotting?  If so I would say leave it and just say that the MacOS version is broken right now.

AsianHam:
You do, but installing GDAL should install it for you.

I've had some edge cases in the past where installing GDAL didn't include
gdal_merge, so I kept a version on hand for simplicity.

mandli:
Oh, so you are just going to chalk this up to the awesome that is GDAL installation woes?  That's fine with me.

rjleveque:
@mandli:  I pointed out above that `gdal_merge.py` is GPL licensed, which is why I suggested removing it and listing it as a dependency instead.

mandli:
@rjleveque I think our replies must have crossed as I never saw your post.  Sorry about that.

PR #283, opened 2021-09-22
Revert "2D Tiff plot option"

rjleveque:
Reverts clawpack/visclaw#282

@AsianHam:  Sorry, I merged #282 without properly testing it and it has at least a couple problems that render it inoperable on some standard examples.

In `frametools.py`:

`    import geoCon as gc`
should be
`    from clawpack.visclaw import geoCon as gc`

Line 183-184:
```
            if plotaxes.plotitem_dict['ITEM1'].plot_type == '2d_gtiff':
                gc.render(framesolns[0], frameno)
```
This is hardwired with the assumption that `plotitem_dict` has a key `ITEM1`.   But if the user calls `plotaxes.new_plotitem` with the `name` argument then the default `ITEM1` is not used.

Moreover, I don't understand why this call is made in the loop on `plotfigure._axesnames` rather than later in the embedded loop on `plotaxes._itemnames` that starts at line 320. Why is this done only for the first item?

But it's not clear to me what the use case is for this, so perhaps the original PR can be fixed up and a new PR issued with more explanation?

Thanks!


rjleveque:
`ClawPlotAxes.new_plotitem` is defined in `visclaw.data` and is where `ITEMn` is specified if `name` is not provided, with `n` incremented for each unnamed item.

In general a user may specify many plot items on a single plot axes, and the items might have different `plot_type`, e.g. perhaps one item with `plot_type='2d_pcolor'` and another with `plot_type='2d_contour'` to put contour lines on top of a pcolor plot, so it doesn't make sense to determine the `plot_type` outside the loop over items.

AsianHam:
I'll take another look and re-PR

Issue #284, opened 2021-10-06
rasterize wavefield by default


Thomas-Ulrich:
I would like to propose a very small enhancement to the plotting routines.
Recently, I generated publication-quality figures with geoclaw.
For that I needed to generate a snapshot of a tsunami simulation in a vectorized format (e.g. svg or pdf).
I only needed the text and axes to be vectorized, not the wavefield.
But using ``plotdata.print_format = 'svg'`` vectorized all data (and lead to a SVG file of 30Mb).
Therefore, I did the following change:
```
diff --git a/src/python/visclaw/frametools.py b/src/python/visclaw/frametools.py
index 9e20d52..dd6f127 100644
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -791,6 +791,7 @@ def plotitem2(framesoln, plotitem, current_data, stateno):
             pcolor_cmd += ", edgecolors=pp['celledges_color']"
         else:
             pcolor_cmd += ", shading='flat'"
+        pcolor_cmd += ", rasterized=True"
 
         pcolor_cmd += ", **pp['kwargs'])"
``` 
Maybe it could be nice to integrate this change?

more info about the rasterization in matplotlib: https://matplotlib.org/devdocs/gallery/misc/rasterization_demo.html

mandli:
Sounds like a good idea to me.

PR #286, opened 2022-04-22
rasterize figures by default

Thomas-Ulrich:
as discussed in https://github.com/clawpack/visclaw/issues/284

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -791,7 +791,9 @@ def plotitem2(framesoln, plotitem, current_data, stateno):
             pcolor_cmd += ", edgecolors=pp['celledges_color']"
         else:
             pcolor_cmd += ", shading='flat'"
-
+        
+        pcolor_cmd += ", rasterized=True"
+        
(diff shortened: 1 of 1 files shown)


mandli:
I tested this a bit as I was not sure what would happen if you had `kwargs = {"rasterized":False}` and it look like it works.  It would be good to document this somewhere although I imagine that there are very few times this would not be the desired behavior for pcolor plots.  Are the other plot types (e.g. `imshow`) also rasterized by default?  Are there other plots we may want to rasterize?

rjleveque:
Thanks @Thomas-Ulrich, this is very useful.  I agree we should document this since it wasn't clear to me at first how to use it, but I think idea is that if the pcolor plot is made this way, you can then do `savefig('filename.svg')`, for example, and it saves an svg file in which the labels use vector graphics but the field plotted is still rasterized.

I tried a quick experiment and the `rasterized=True` option can also be added to `imshow` commands with similar advantage, although the svg file made without this option is already much smaller than one made with `pcolor`.

mandli:
I definitely think we should default to rasterization when it makes sense.

PR #287, opened 2022-04-24
Update ianimate to work with Matplotlib only (no more JSanimation).

ketch:
Today I found that this doesn't work (running https://github.com/clawpack/apps/blob/master/notebooks/pyclaw/Acoustics-1D.ipynb).  I've revised it to stop using visclaw.JSAnimation.  I think we should probably delete that, as it likely is broken in other contexts.

### DIFF
--- a/src/python/visclaw/ianimate.py
+++ b/src/python/visclaw/ianimate.py
@@ -6,14 +6,17 @@ from __future__ import absolute_import
 
 def ianimate(frame_list,plotdata=None,ivar=0,varname=None,**kargs):
     """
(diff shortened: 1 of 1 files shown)


rjleveque:
Yes, we stopped using JSAnimation in other places (plotpages.py, animation_tools.py) and notebooks like [this one](http://www.clawpack.org/gallery/_static/apps/notebooks/visclaw/animation_tools_demo.html), but I missed this usage.

I agree we should remove the JSAnimation subdirectory since it's no longer used.  I can do a PR for that.

PR #291, opened 2023-08-30
Add more attributes to some plot classes to simplify things commonly done

rjleveque:
I added new attributes to `ClawPlotFIgure` and `ClawPlotAxes` to simplify doing various things in `setplot.py`, all of which could be done using `kwargs` and/or `afteraxes` commands, but these are things I do so frequently that I think they should be made simpler.  

I think it is all backward compatible, it just makes some things easier.

The figure below shows a new version of a plot from the standard chile2010 example, using this modified setplot.py that takes advantage of some of the new attributes: https://github.com/rjleveque/geoclaw/blob/plot_updates/examples/tsunami/chile2010/setplot.py

Note in particular:
- white background
- axes labeled Longitude and Latitude with appropriate fontsize
- the title and tick labels also have the fontsize set more easily
- title has time in hour:minute:second notation
- the colorbar has arrows at top and bottom (`extend = 'both'`) to show that waves may be larger than limits of colorbar

chile2010_frame5

Compare to http://www.clawpack.org/gallery/_static/geoclaw/examples/tsunami/chile2010/_plots/frame0005fig0.html

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -692,12 +692,12 @@ class ClawPlotAxes(clawdata.ClawData):
         self.add_attribute('xticks_kwargs', {}) # e.g. to set ticks,rotation
         self.add_attribute('yticks_fontsize', None) 
         self.add_attribute('yticks_kwargs', {}) # e.g. to set ticks
-        self.add_attribute('x_label', None) # label for x-axis
-        self.add_attribute('y_label', None) # label for y-axis
-        self.add_attribute('x_label_fontsize', None)
-        self.add_attribute('y_label_fontsize', None)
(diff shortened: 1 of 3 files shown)


rjleveque:
I decided to add fontsize attributes directly to title, `title, xlabel, ylabel, xticks, yticks`, to require less use of kwargs for this common situation, and updated the [chile2010 example](https://github.com/rjleveque/geoclaw/blob/plot_updates/examples/tsunami/chile2010/setplot.py) to use this, and also to illustrate how `xticks` locations and rotation can be specified in `xticks_kwargs`
chile2010_frame5a
.

rjleveque:
These seem to be working well.

PR #294, opened 2023-11-01
update geoplot.py to handle 1d or 2d q array, and also eta = q[-1,...]

rjleveque:
Use ellipses q[0,...] for depth, which is q[0,:] in 1d or q[0,:,:] in 2d. 

Use eta = q[-1,...] to work in 1d or 2d, and also with 2d bouss where q will have additional components for hu and hv correction terms.  Assumes that eta is always the final component of the q array read in from geoclaw output.

### DIFF
--- a/src/python/visclaw/geoplot.py
+++ b/src/python/visclaw/geoplot.py
@@ -8,7 +8,6 @@ import warnings
 from clawpack.visclaw import colormaps
 from matplotlib.colors import Normalize
 from clawpack.geoclaw import topotools
-# from six.moves import range, zip
 
 
 # Colormaps from geoclaw
--- a/src/python/visclaw/geoplot.py
+++ b/src/python/visclaw/geoplot.py
@@ -2,15 +2,13 @@
 Useful things for plotting GeoClaw results.
 """
 
-from __future__ import absolute_import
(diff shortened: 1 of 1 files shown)


src/python/visclaw/geoplot.py
```diff
@@ -2,15 +2,13 @@
...
-from __future__ import absolute_import
-from __future__ import print_function
 import numpy
 import warnings
 
 from clawpack.visclaw import colormaps
 from matplotlib.colors import Normalize
 from clawpack.geoclaw import topotools
-from six.moves import range, zip
+# from six.moves import range, zip
```
mandli, line 11: Might as well just remove this.

mandli (approved):
Looks good to me other than the minor comment.  Do you want to also include an example in this PR or just wait?

src/python/visclaw/geoplot.py
rjleveque: done.

rjleveque:
@mandli: Not sure what sort of example you mean since we don't have any examples in visclaw repo now.  For old examples in geoclaw it should work the same as before, I hope.

mandli:
Sorry, was looking at too many PRs.

PR #297, opened 2023-11-14
Add ability to pass norm and alpha values to imshow

kbarnhart:
I've found it useful to be able to specify a norm object to plotitem for 2d_imshow

e.g., 

```python
plotitem.imshow_norm = mpl.colors.LogNorm(0.001, 10, clip=True)
```
 
Similarly, I've found it useful to provide an alpha value to make an imshow transparent (e.g., on top of a hillshade). 

This PR adds these two features (I'll make an associated PR for the docs shortly).

### DIFF
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -845,6 +845,7 @@ class ClawPlotItem(clawdata.ClawData):
                 self.add_attribute('imshow_cmin',None)
                 self.add_attribute('imshow_cmax',None)
                 self.add_attribute('imshow_norm', None)
+                self.add_attribute('imshow_alpha', None)
 
 
             elif plot_type in ['2d_contour', '2d_contourf']:
--- a/src/python/visclaw/data.py
+++ b/src/python/visclaw/data.py
@@ -844,6 +844,7 @@ class ClawPlotItem(clawdata.ClawData):
                 self.add_attribute('imshow_cmap',colormaps.yellow_red_blue)
                 self.add_attribute('imshow_cmin',None)
                 self.add_attribute('imshow_cmax',None)
+                self.add_attribute('imshow_norm', None)
 
 
             elif plot_type in ['2d_contour', '2d_contourf']:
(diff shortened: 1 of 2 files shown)


kbarnhart:
This is not quite ready to be considered because the norm is not being correctly reflected in the colorbar. Will note here when it is ready for further consideration.

mandli:
There are some tools to do this but I am not sure that they match what you would like.  The norms are always more difficult to work with so it also may be something we would put into an example or FAQ if the implementation is too difficult.

kbarnhart:
thanks @mandli - If I get it to work I will make an example. I got it to work before... so 🤞 I can get it to work again. 

point me towards the tools? If they are sufficient, I'd happily migrate.

rjleveque:
@kbarnhart:  It seems like adding it as you do here should work, and would be a good addition. I'm not sure offhand why it's not reflected in the colorbar.

 In looking at the code I see that we allow setting `plotitem.kwargs` for `pcolor` plots and these are appended to the plot command, but not for `imshow`.  I'm not sure why not, but that might also be nice to allow for users who want to pass in other arguments we don't explicitly support.

kbarnhart:
After a bit of digging, I concluded that the reason I thought it was not working was because no colorbar ticks were being draw (thus it was not possible to evaluate if the colorbar norm was working as expected). The following example works as expected for the chile2010 case:

```python
    from matplotlib.colors import SymLogNorm
    #.... many missing lines
    plotitem = plotaxes.new_plotitem(plot_type='2d_imshow')
    plotitem.plot_var = geoplot.surface_or_depth
    plotitem.imshow_cmap = geoplot.tsunami_colormap
    plotitem.imshow_norm = SymLogNorm(vmin=-2, vmax=2, linthresh=0.02)
    plotitem.imshow_alpha=1.0
    plotitem.add_colorbar = True
    plotitem.colorbar_ticks = [-2, -.2, -0.02,0,0.02, 0.2, 2]
    plotitem.colorbar_tick_labels = [-2, -.2, -0.02,0,0.02, 0.2, 2]
    plotitem.amr_celledges_show = [0,0,0]
    plotitem.patchedges_show = 1
```

@rjleveque and I just discussed whether additional kwargs might ever need to be passed to imshow (besides the two I've just added) and concluded that no other kwargs would make sense to pass. Therefore, I've not added additional kwargs. 

I think that this pr and https://github.com/clawpack/doc/pull/227 are ready to be considered.

PR #298, opened 2023-12-08
Various updates to Matlab code in VisClaw

donnaaboise:
These are mostly minor updates to fix style issues. 

* Fix indenting
* Remove some unnecessary commas
*  Update a few deprecated Matlab calls

A more substantial fix is to remove masking in ForestClaw when drawing contour lines.  Patch masking (needed in AMRClaw, but not in ForestClaw) was being called in ForestClaw to mask contour lines on "covered" patches.  This basically was hiding the entire patch, leaving holes in the patch in some cases.

### DIFF
--- a/src/python/visclaw/colormaps.py
+++ b/src/python/visclaw/colormaps.py
@@ -16,6 +16,9 @@ import numpy
 import matplotlib.colors as colors
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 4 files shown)


donnaaboise:
These are all changes that are mostly cosmetic and should have little affect on the API for the Matlab routines.

rjleveque:
Thanks @donnaaboise!

Issue #307, opened 2024-09-27
`make plots` is slow


giboul:
Hi! I' new here but I'd like to make a suggestion.

The `make plots` command can be frustratingly long when debugging. A faster solution would be to use `Iplotclaw` but I did not appreciate having to use the terminal at the same time (and having to minimize windows because I have a small screen).

So I wrote a small script making use of the [`plotframe(..., refresh=True)`](https://github.com/clawpack/visclaw/blob/master/src/python/visclaw/frametools.py#L42) function and matplotlib's [mpl_connect](https://matplotlib.org/stable/api/backend_bases_api.html#matplotlib.backend_bases.FigureCanvasBase.mpl_connect) to change frames with just the arrow keys. I ended up adding numerical input (press 49 then press enter to show frame 49) and a FuncAnimation when passing the `FPS=<some_frames_per_second_value>` argument to `make iplot`.

Could this be a good addition to visclaw? Here my code

```Python
from pathlib import Path
from argparse import ArgumentParser
from matplotlib import pyplot as plt
from matplotlib.animation import FuncAnimation
from clawpack.visclaw.Iplotclaw import Iplotclaw
from clawpack.visclaw.frametools import plotframe


def interactive_plot(fps=0, **kwargs):
    """Take advantage of Iplotclaw to make an interactive plotting session."""
    ip = Iplotclaw(**kwargs)
    counter = dict(current=0, max=0, right=1, left=-1, up=1, down=-1)
    counter["max"] = len(list(Path(ip.plotdata.outdir).glob("fort.q*")))

    def update(i):
        plotframe(i, ip.plotdata, simple=ip.simple, refresh=True)
    update(0)

    def arrow_key(event):
        """If pressed key is in counter, increment index and update plot."""
        if event.key in counter:
            counter["current"]=(counter["current"]+counter[event.key])%counter["max"]
            update(counter["current"])

    frame = dict(n="")
    def to_frame(event):
        if event.key.isnumeric():
            frame["n"] += event.key
        elif event.key == "enter":
            counter["current"] = int(frame["n"] or 0)
            frame["n"] = ""
            update(counter["current"])


    if fps:
        anim = FuncAnimation(plt.gcf(), update, frames=counter["max"], interval=1e3/fps)
        return anim
    else:
        for fn in plt.get_fignums():
            plt.figure(fn).canvas.mpl_connect('key_press_event', arrow_key)
            plt.figure(fn).canvas.mpl_connect('key_press_event', to_frame)


[...]

PR #308, opened 2024-10-01
Mpl connect

giboul:
As per issue [`make plots` is slow](https://github.com/clawpack/visclaw/issues/307), this is an attempt to use `mpl_connect` in parallel with `cmd` to avoid having to navigate between the CLI and the figure. A later addition (idk how to do it yet) could be to create and save a FuncAnimation when `--fps` or `--fname` are passed to the command.


src/python/visclaw/ipyclaw.py
```diff
@@ -6,15 +6,23 @@
...
+from argparse import ArgumentParser
+from clawpack.visclaw.Iplotclaw import Iplotclaw
 
-def run_iplotclaw(setplot='setplot.py',outdir='./_output'):
-    ip = Iplotclaw.Iplotclaw(setplot=setplot,outdir=outdir)
+def run_iplotclaw(**kwargs):
+    ip = Iplotclaw(**kwargs)
     ip.plotloop()
 
+def parse_args():  # TODO: ipython cannot parse custom arguments
```
mandli, line 16: We should be able to call this using Python so I am not sure I understand why IPython is an issue.

src/python/visclaw/ipyclaw.py
```diff
@@ -6,15 +6,23 @@
...
+    parser.add_argument("--fps", default=None, type=float, required=False)
+    args = parser.parse_args()
+    return args
+
 if __name__ == '__main__':
-    if len(sys.argv) > 1:
-        run_iplotclaw(*sys.argv[1:])
-    else:
-        run_iplotclaw()
+    run_iplotclaw(setplot="setplot.py", outdir="./_output")
```
mandli, line 28: Is this just a placeholder until the command line parsing can be used?

src/python/visclaw/Iplotclaw.py
```diff
@@ -11,6 +11,7 @@
 """
```
mandli, line 12: All these imports, unless absolutely necessary, should be directly imported rather than imported into the global namespace.  In other words `import matplotlib.pyplot as plt` instead of `from matplotlib import pyplot as plt`.  Also, you do not need to check `sys.modules` for `matplotlib`, you can just import it and set the values.  I am not sure if `matplotlib.interactive` needs to be called before `import matplotlib.pyplot` is done though this would not matter in either case.

mandli (commented):
Just a couple of quick comments on what you have implemented so far.

src/python/visclaw/ipyclaw.py
giboul: Ok I'm just a noob, I've just found out about the `-i --` flag. Removing it now.

src/python/visclaw/Iplotclaw.py
giboul: The `sys.modules` check what not of my doing but I agree to the changes.

src/python/visclaw/ipyclaw.py
giboul: Yes, `**parse_args().__dict__` is now passed instead.

mandli:
Beyond the last comment about defaults for the CLI I this works for me.  If you make that last change I think we can merge this.

src/python/visclaw/ipyclaw.py
```diff
@@ -6,15 +6,21 @@
...
+    parser.add_argument("--fps", default=None, type=float, required=False)
+    args = parser.parse_args()
+    return args
+
 if __name__ == '__main__':
-    if len(sys.argv) > 1:
-        run_iplotclaw(*sys.argv[1:])
-    else:
-        run_iplotclaw()
+    run_iplotclaw(**parse_args().__dict__)
```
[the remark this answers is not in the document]
giboul: Right , I just added the defaults (and got carried away with a messy commit), should be all good.

giboul:
Well, thanks for being so welcoming! I'm happy to contribute, this my first PR ever :D

mandli:
Thanks @giboul for you work on this.  I started thinking about other things we could have key bindings to but lets leave that as future work/suggestions.

PR #309, opened 2024-10-21
Make plotpages respect file format specification.

ketch:
This fixes a bug that caused visclaw to look for fort.q* files even when the user specified the file format as petsc. Now it correctly looks for claw.ptc* files.


mandli:
Looks good to me but I have not been able to test it yet.

ketch:
If there aren't any objections I'll go ahead and merge this tomorrow.

rjleveque:
Works for me, thanks @ketch.

PR #313, opened 2025-03-04
Remove Cygwin Code

mandli:
This PR removes old code that used to be needed when used with Cygwin.  Clawpack has not directly supported Cygwin for a bit and there are better [alternatives](https://learn.microsoft.com/en-us/windows/wsl/).

This was removal was caused by a bunch of errors such as 
```sh
$CLAW/visclaw/src/python/visclaw/plotpages.py:2877: SyntaxWarning: invalid escape sequence '\c'
  plotdir = 'C:\cygwin' + plotdir
```
 from newer versions of Python that do not allow `\c` characters.  This may be able to be modified but I do not have a cygwin installation to test from and this seems as good as any a time to remove the support.

### DIFF
--- a/src/python/visclaw/frametools.py
+++ b/src/python/visclaw/frametools.py
@@ -1349,18 +1349,6 @@ def printframes(plotdata=None, verbose=True):
 
     rootdir = os.getcwd()
 
-    # annoying fix needed when EPD is used for plotting under cygwin:
-    cw_path = r'C:\cygwin'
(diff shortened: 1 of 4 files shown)


ketch:
This seems good to me.  I think there is a block of code in frametools.py starting at line 1352 that should also be removed.  I didn't find any others, but I didn't check very carefully.

mandli:
Ah, you were right, missed that one, although, for whatever reason, it was not raising a warning.

PR #315, opened 2025-05-05
add plotdata as argument to plotclaw

rjleveque:
Needed for some new multiprocessing code to be added to clawutil for doing parameter sweeps.

Should be backward compatible since default argument is `None`.


mandli:
See clawpack/clawutil#180 for a discussion.

Issue #317, opened 2025-06-19
Slowdown When Using Multiple Threads to Plot


mandli:
I have noticed that when `OMP_NUM_THREADS` > 1 that the last images take a much longer time to plot than with `OMP_NUM_THREADS == 1`.  Everything else until that last set plots quickly as expected and I do not think that things are waiting around as the single thread plotting is faster overall.

rjleveque:
I'm not sure what's going on but I don't think it's because the last png file takes longer to make, I think there's some other stuff going on after they are all created that I haven't sorted out since the logic of this module is very convoluted in the way we implemented multithreading.  I did notice something strange and added a comment to #309.  I will try to dig into this further...

mandli:
That was my thought too, but have yet to track it down.  Is strange that it does not seem to happen without the parallel part.  Maybe there's a better way to do that in general.

PR #318, opened 2025-08-01
Shorten poll interval for process checking to 0.5 seconds

mandli:
This seems to address #317.  When multiple processes are used to do plotting a poll timer is used so that the processes are not checked constantly.  This used to be set to 5 seconds, but it turns out this may have been a bit too long.  I am not entirely sure what to set this to, but an order of magnitude less seems a good start as it removes this particular bottleneck.


rjleveque:
looks good and works for me, thanks!

PR #319, opened 2025-08-16
animation_tools fix to sort filenames for making animation

rjleveque:
In animation_tools.read_images, so the images are in the right order, assuming zero-padded frameno is part of the fname_pattern.


mandli:
Was this just working on luck before?  Not ideal I suppose.

rjleveque:
Probably just luck, but also maybe nobody has been using this function until @mjberger tried it the other day.

